Issue #39, opened 2015-02-13
Listening on the same port used for sending


dmoranj:
I'm currently involved in the development of a OMA Lightweight M2M library, based on node-coap (you can find it [here](https://github.com/telefonicaid/lwm2m-node-lib) and I'm struggling with one aspect of the specification: it requires the server to communicate with the devices **using the same IP and Port** used in the registration message (a POST COAP message to the server). That IP and port is used then for bidirectional communications: the server can send COAP Messages to that address and that is the address that is expected to be used to initiate messages from the device side as well.

I'm trying to overcome this issue using a customized Agent for every request, bound to the same 5683 port where the device listens for requests originated from the server. The problem here is that this situation generates two listening handlers: the one set up in the createServer() function, and the listener for the "response" event from the request. The problem arises when a request is sent using port 5683, and the response arrives to the port with two listeners.

This wouldn't be a problem in case both handlers were called, as I could simply ignore answers in the server handler (e.g.: identified by the lack of method), but that's not the case: the answer to the request arrives exclusively to the server handler, thus leaving the response listener blocked forever.

I know this scenario may be too specific for the Lightweight M2M protocol, but I would appreciate any idea for dealing with this problem, as I'm a bit stuck with it, including any contribution I could make to the library to support this weird case.

dmoranj:
Working on this problem I've realized that the specification doesn't quite specify that *all the requests from the client* should use the same port, but that *the client should listen in the same IP and port** of the registration message. This opens makes the solution far easier: I just need read the port from the outgoing socket and, once I receive the response from the registration, open a new socket to listen in that same port.

I have a problem implementing this approach, though: I do not how to extract the port used in a request sent to the COAP Server from the req or res. I've managed to solve it temporary by modifying the OutgoingMessage class to add the address of the socket to the `_rsinfo` attribute, so it can be read later. What I wonder is: Am I missing a better option to extract that port? Will it be interesting to contribute that feature to the library, to help retrieving the address information, or does it have another way to retrieve it that I haven't seen?

mcollina:
Ahum, I think that the agent and the server might get to share the same port to support Lightweight M2M. However, I have no bandwidth to dedicate time to such a refactoring. However, if you would like to submit a PR for it, I will try my best to help. Moreover, this is an OPEN Open Source project :), and you will join us in maintaining this library.

dmoranj:
[...]

PR #40, opened 2015-02-16
Original address

dmoranj:
Add a property in the response listener for the requests to allow the caller to retrieve information about the socket used in the communication (port, address and protocol). This addition was motivated by the need of identifying the device original address in COAP-based OMA Lightweight M2M protocol as explained in issue #39 (originally the need was to use the same socket for listening and sending requests, but having information about the registration socket might lead to avoid the sharing, as explained in a comment).

I choose to expose the information as part of the `rsinfo` object (that was already passed to the IncomingMessage instance creation) and expose the latter in a `_rsinfo` internal attribute of the response, but I'm not 100% confident this is the best option. It covers the functionality specified in the issue, though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -362,6 +362,15 @@ following additional methods, events and properties.
 
 Closes the stream.
 
+#### message.rsinfo
+
+The sender informations, as emitted by the socket.
+See [the `dgram` docs](http://nodejs.org/api/dgram.html#dgram_event_message) for details
+
+#### message.outSocket
+
+Information about the socket used for the communication (address and port).
+
 -------------------------------------------------------
 <a name="observewrite"></a>
 ### ObserveWriteStream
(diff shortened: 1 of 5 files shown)


mcollina:
I'm ok with adding that visibility of the local address, however, I am not sure that `_rsinfo` is the right place for it. I would likely expose the rsinfo as `rsinfo` as it would become part of the external interface, and document that in the README. But I would also keep that untouched (as from the UDP socket itself) and add one more property.

Also, you would need to add them to https://github.com/mcollina/node-coap/blob/master/lib/observe_read_stream.js.

dmoranj:
Ok. I'll expose the `rsinfo` as is and the outgoing socket as an `outSocket` one, and add the docs for both.

dmoranj:
Information about both sockets exposed as external attributes and added to their respective sections in the documentation.

mcollina:
@dmoranj are you ok to become a collaborator of this project? Have a look at https://github.com/mcollina/node-coap/blob/master/CONTRIBUTING.md and let me know!

dmoranj:
Sorry, didn't read this message until now. Yes, I would like to become a collaborator (I guess I'll be working extensively with COAP in the next months, so I hope I'll have time to contribute).

Issue #15, opened 2013-12-06
Agent is not closed in case of errors


mcollina:
If an `error` event happen in a request, the agent is not closed properly and the socket remains open.

PR #41, opened 2015-02-16
Clean up agent socket on error

dmoranj:
Clean up the agent in case of error. I met with some problems due to open sockets when there are errors in the connection, and this patch seem to solve them. Tries to address issue #15

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -301,7 +301,10 @@ Agent.prototype.request = function request(url) {
     }
   }
 
-  req.sender.on('error', req.emit.bind(req, 'error'))
+  req.sender.on('error', function(error) {
+    that._cleanUp();
+    req.emit('error', error);
+  })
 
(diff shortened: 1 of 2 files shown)


mcollina:
I am not sure about this: I think the agent should not close if there is an error in the connection. However, the default one should do, i.e. `coap.request()`. It is possible that an agent is used to do more than one request at the same time, and killing one will kill the other too.

dmoranj:
This was motivated by some really difficult to debug problems with the default `coap.request()` version of the client, when the socket was reused after a client error (listeners from older executions seem to be called instead of newer ones). Cleaning the sockets after the error was enough to get rid of all the problems.

Anyway, I have solved them using a specific new agent in the client, but it seems to me this may cause some trouble in the future. I will close this PR and try to better understand the issue (and provide a clear case of the error, without the rest of my code ofuscating it) before proposing a new one.

Issue #43, opened 2015-02-23
Provide an easy way to mark a response as RESET


dmoranj:
This behavior can be currently achieved in the following way: 
```
server.on('request', function (req, res) {
    console.log('This is it: ' + req.payload.toString());
    res._packet.reset = true;
    res._packet.ack = false;
    res.end();
});
```
It seems there isn't any direct way to achieve the same result with the public API. The RESET behavior can be useful to implement some behaviors, as COAP Ping (a RST response to an empty confirmable body, as specified in the COAP specification).

mcollina:
I'm definitely :+1: of having this in this library. Would you like to send a PR to offer a `res.reset()` thing?

On the other end, how this will be exposed to the agent on the client side?

dmoranj:
No problem, I'll try to implement it tomorrow morning. I understand, from the client point of view, a Reset is a response, but with a proper flag indicating is not an ACK, but a RST coming, but I'm not really confident about that (maybe a different event should be emitted?). 

I also doubt if it would be nice to directly implement the COAP Ping mechanism in the server (maybe with a flag to enable/disable it), or if it should be left to implementors of COAP Servers to replicate that behavior in their servers.

mcollina:
I would rather have a module that does the ping logic that on top of this one.

mcollina:
Fixed by #44.

PR #44, opened 2015-02-24
Add reset method

dmoranj:
Add a Reset method to the Outgoing message, so a COAP RST message can be returned to the caller.

### DIFF
--- a/test/server.js
+++ b/test/server.js
@@ -16,7 +16,7 @@ var coap      = require('../')
   , sinon     = require('sinon')
   , params    = require('../lib/parameters')
 
-describe('server', function() {
+describe.only('server', function() {
   var server
     , port
     , clientPort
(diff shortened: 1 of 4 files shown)


mcollina:
You have to add it to https://github.com/mcollina/node-coap/blob/master/lib/observe_write_stream.js too.

dmoranj:
Added method to the Observe Write Stream.

dmoranj:
This PR address issue #43

test/server.js
```diff
@@ -215,6 +215,23 @@ describe('server', function() {
     })
   })
 
+  // TODO
```
mcollina: why is there a todo?

test/server.js
```diff
@@ -16,7 +16,7 @@ var coap      = require('../')
   , sinon     = require('sinon')
   , params    = require('../lib/parameters')
 
-describe('server', function() {
+describe.only('server', function() {
```
mcollina: why `only()`?

mcollina:
Can you also add a test for the `ObserveWriteStream`? Thanks!

mcollina:
Also, add yourself to https://github.com/mcollina/node-coap#contributors :).

test/server.js
dmoranj: Because I forgot to remove it :D Same goes for the "TODO". God bless Code Reviews.

dmoranj:
The test labeled "'should send a \'RST\' to the client if the msg.reset() method is invoked'" address the reset() function for the ObserveWriteSrteam.

dmoranj:
Commit  3b6565c fixes the `only()` and "TODO"

mcollina:
Released!

PR #45, opened 2015-03-03
Server shut down on unexpected option

dmoranj:
With the current behavior, an incoming request with an unknown 'Content-Format' option generates an unexpected exception that makes the server crash. I've captured the exceptions in the message handling, to return an error to the user, indicating that the request could not be handled.


mcollina:
Thanks!!

PR #46, opened 2015-03-04
Single socket

dmoranj:
Adds two features to the library:
* Add the `proxy` option to the server creation, so the server will behave like a COAP Proxy: if a request arrives containing the Proxy-Uri Option, instead of treating it like a normal request, it is forwarded to the destination indicated in the Option, removing the Proxy-Uri option from the package.
* Add the `proxyUri` option to the request options, so it can be directly indicated in the request creation (just adds the Proxy-Uri option to the final request).

In order to implement the server option, an important refactor in the request handler was done. I extracted the request handler function to a top-level function, and decomposed the process into a `parseRequest` and a `handleServerRequest` function. This functions were composed as a concatentation of middlewares with the signature `function(request, next)` being the `request` parameter a data object shared among all the middlewares processing a request and the `next()` function a callback to handle the pass to the next middleware. If the `next()` function is executed with a non-zero argument, the request processing stops and the final request processor is executed. If the parameter to the function was the string `end` the request is considered to have been processed correctly and no further action is done. Otherwise, the parameter is considered an error description and a `5.00` error is returned to the client. The list of middlewares for a server is stored in the `_middlewares` internal attribute of the server.

For the proxy features, two other middlewares were added: 
* `proxyRequest`: handles incoming requests containing the `Proxy-Uri` option. It removes the option and resend the request to the destination.
* `handleProxyResponse`: handles responses to proxied requests, using an internal `_proxiedRequests` object, matching the incoming requests using the `token` parameter of the request (for proxied requests, if no token is provided a new 8-byte length token is generated).

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ __node-coap__ is a client and server library for CoAP modelled after the `http` 
   * <a href="#intro">Introduction</a>
   * <a href="#install">Installation</a>
   * <a href="#basic">Basic Example</a>
+  * <a href="#proxy">Proxy features</a>
   * <a href="#api">API</a>
   * <a href="#contributing">Contributing</a>
   * <a href="#licence">Licence &amp; copyright</a>
@@ -103,6 +104,23 @@ server.listen(function() {
   req.end()
 })
 ```
+<a name="proxy"></a>
+## Proxy features
+The library now comes with the ability to behave as a COAP proxy for other COAP endpoints. In order to activate the
+proxy features, create the server with the `proxy` option activated.
+
+A proxy-enabled service behaves as usual for all requests, except for those coming with the `Proxy-Uri` option. This
+requests will be redirected to the URL specified in the option, and the response from this option will, in turn,  be
+redirected to the caller. In this case, the proxy server handler is not called at all (redirection is automatic).
+
+You can find an example of how this mechanism works in `examples/proxy.js`. This example features one target server
+that writes all the information it receives along with the origin port and a proxy server. Once the servers are up:
+
+- Ten requests are sent directly to the server (without reusing ports)
+- Ten requests are sent through the proxy (without reusing ports)
+
+The example shows that the target server sees the last ten requests as coming from the same port (the proxy), while the
+first ten come from different ports.
 
 <a name="api"></a>
 ## API
--- a/README.md
+++ b/README.md
@@ -193,7 +193,7 @@ The constructor can be given an optional options object, containing one of the f
 * `type`: indicates if the server should create IPv4 connections (`udp4`) or IPv6 connections (`udp6`). Defaults
   to `udp4`.
 * `proxy`: indicates that the server should behave like a proxy for incoming requests containing the `Proxy-Uri` header.
-  Defaults to `false`.
+  An example of how the proxy feature works, refer to the example in the `/examples` folder. Defaults to `false`.
 
 
 #### Event: 'request'
--- a/README.md
+++ b/README.md
@@ -152,6 +152,9 @@ If it is an object:
   * [`Agent`](#agent) object: explicitly use the passed in [`Agent`](#agent).
   * `false`: opts out of socket reuse with an [`Agent`](#agent), each request uses a
     new UDP socket.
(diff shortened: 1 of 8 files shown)


mcollina:
Thank so much. I think this is a great solution, and a feature we should support in node-land. 
Big question: can this be implemented on top of node-coap somehow? It seems it needs access to low-level stuff, but it might be good to know.

Generally, I would prefer adding less features as they require more maintenance.

Regarding the code, I think the only wrong piece is how the `next()` function is called. Performance-wise, a function should always be called always with the same types.

Also, can you please add an example that demonstrate this, maybe using a couple of processes.

dmoranj:
Regarding the examples, I think is a good idea, I'll try to add it. I think it will be difficult to implement on top, but I can think about it (maybe we could refactor and extract common functionalities between proxy and server and separate the proxy part).

I'm have also concerns with the call to `next()`. I'll check out how the Connect middleware system is implemented in case it gives me any ideas to improve it.

I'll be at least a week without working on this, as I'm really, really busy this week, but I'll try to address all this concerns this weekend.

mcollina:
Don't worry and take your time to do this :). You are the main driver for this at this point in time anyway!

dmoranj:
Regarding the `next()` function problem, Is your corncern related with the difference between types like this example shows?: 
```
function handleServerRequest(request, next) {
  try {
    request.server._handle(request.packet, request.rsinfo)
    next('end')
  } catch (err) {
    next(new Buffer(err.message))
  }
}
``` 
Or is it related with how they are bound in code like this one?:
```
```

mcollina:
The first case. From a V8 point of view, the first one is not optimizable.

dmoranj:
I have made some changes to address your concerns and some of mine.

- I have changed the type of the next() function to always use a Buffer. I also added a boolean flag to check whether to stop the middleware chain or not, instead of a comparison with the Buffer content.

- I added a test for error conditions in proxied requests. This raised some problems with how the errors where returned to the caller (I added the messageId and token to the error response).

- I added a proxy example, indicating how a server behaves with and without proxy.

The Travis seems to be having some problems with the commits, I'm checking that (local tests seem to work without any problem). It may be related with one test that fails occasionally (the one counting the number of replies in a certain number of miliseconds); those tests based on time are usually a bit weak, o maybe is a false positive.

dmoranj:
Seems to be a problem with the error text being version dependant (ENOTFOUND seems to print the hostname in earlier versions). I'll modify the check and recommit.

mcollina:
I think we are close :). I'm happy to merge this. Is there something more to add?

Does this support perfectly lwm2m?

dmoranj:
I hope so :) but I will have to work with the library and some standard OMA tools to make sure it works as it is expected. Either way, I think the proxy feature is a nice addition. 

The only thing that may be missing concerning this particular feature is some more documentation (maybe a paragraph explaining briefly what the proxy thing is about).

mcollina:
Good, add it to the docs :D.

We might think later on to split this library into multiple low-level modules that provides blockwise, observe, etc..

dmoranj:
Done. Added a paragraph about the proxy in the documentation :)

dmoranj:
I've made some tests with the new features, and it seems to work now; but I'm facing other different problems. Shall we close this PR and put the next issues in a different one?

dmoranj:
I have added a fix for the case where the content format is bigger than a byte, as the code was clipping the value, so making some mistakes in the communication.

mcollina:
Before releasing I removed async and passed to [fastseries](http://npm.im/), which is way faster as the name suggest :). I also removed all those async.apply, which creates a lot of unneeded functions for handling a request.

dmoranj:
:( Lots of tests have stopped working for realease 0.10.1. Do they work in your local machine? The changes also seem to have broken the changes in my LWM2M library (it was working with my forked branch, but now even the basic functionality is broken). Has the basic functionality changed in any way with your changes?

mcollina:
all tests pass here and on travis too https://travis-ci.org/mcollina/node-coap/builds/58714533

dmoranj:
Ok. I had some processes using the same ports. Once closed, it works fine (the tests, there's still something broken in the LWM2M library, but that's my problem alone :P ).

PR #54, opened 2015-04-16
Fix server functionality in proxies

dmoranj:
With the last changes (the introduction of "fastseries"), the standard server functionality for servers with the proxy feature activated was broken, i.e: if a server is created with the `proxy` option, and a request is sent without the `proxyUri` option, the usual handler should be called, but, with the last modifications, a `missing request` error is received instead. This PR address that case.

### DIFF
--- a/lib/middlewares.js
+++ b/lib/middlewares.js
@@ -74,7 +74,7 @@ function handleProxyResponse(request, next) {
     delete request.server._proxiedRequests[request.packet.token.toString('hex')]
     next(null)
   } else {
-    next(new Error('missing request'))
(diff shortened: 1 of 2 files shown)


mcollina:
Thanks! :).

dmoranj:
I saw you created the 0.10.2 release, but it still doesn't download from the NPM registry. Is it a cache problem? Did you publish the release?

mcollina:
done.

Il giorno gio 16 apr 2015 alle ore 16:49 Daniel Moran <
notifications@github.com> ha scritto:

Issue #56, opened 2015-05-14
Proxy blocks observe responses


dmoranj:
Identified by @ClaesNilsson

If a GET observe request is sent through the proxy then only the first response of the request is received.

dmoranj:
I've tried to reproduce with tests (view PR #57) and it won't reproduce. Can you provide a simple example to reproduce it? Can you give us more information about the version of your environment or any other factors that may be affecting the test? Thanks.

ClaesNilsson:
Ok, I will provide a test case. However, I am just about to go on a business trip so may not be able to provide the test case until Friday this week.

dmoranj:
Ok, when you want. Once we have the test case we will start working on solving it.

dmoranj:
Also find the problem: the proxy removes the request entry once it has been answered, but that's wrong in the case of the Observer requests. I'll change that to add an exception.

Issue #55, opened 2015-05-14
Handler not called when proxyUri not present


dmoranj:
Identified by @ClaesNilsson

If a server received a request without a proxyUri field the handler of the server should be called instead of forwarding to another target server. This does now work, no handler is called in this case.

dmoranj:
@ClaesNilsson, checking this bug, it seems the scenario is covered by the 'with a non-proxied request should call the handler as usual' test in the `proxy.js` test file, so the basic case seems to be working. Can you provide a simple example to reproduce the bug? Thanks.

ClaesNilsson:
Ok, I will provide a test case. However, I am just about to go on a business trip so may not be able to provide the test case until Friday this week.

dmoranj:
Found the problem with your examples, thank you. The problem was the createServer() method did not pass along the "listener" method, so it was not subscribed to the 'request' event. I'll propose a PR ASAP.

PR #59, opened 2015-05-22
Added proxy tests referring:

ClaesNilsson:
https://github.com/mcollina/node-coap/issues/55
https://github.com/mcollina/node-coap/issues/56


mcollina:
Can you please format the tests so that they follows the structure of the of the others? So they will be picked by travis, too!

dmoranj:
Maybe this PR is not needed anymore, or maybe it can be added in an examples folder. I've already created a small test that covers #55 and I'm in process of doing the same with #56.

ClaesNilsson:
Thanks Daniel! 
So do you think that it adds any value if I add my tests to the examples folder?

dmoranj:
It's a good question, I don't know what Matteo thinks about this: should we give as much examples as possible or should we give just small examples one per feature? I think the most interesting thing about the example was to show the bugs themselves, and that functionality is already exposed in the tests, but I prefer to listen to Matteo's opinion on the subject.

mcollina:
I think it's ok to have some integration tests, added to ensure bugs like
this one. Let's put them in a folder like 'test/integration', so we know
what we are talking about.
Il giorno lun 25 mag 2015 alle 15:16 Daniel Moran <notifications@github.com>
ha scritto:

ClaesNilsson:
Just to make sure that there is no misunderstanding. I create a subfolder under 'test' (or 'exampels'?) called 'integration' and under that folder with my tests called 'ProxyTests'?

mcollina:
why not just putting them in 'test/integration'?
Il giorno mar 26 mag 2015 alle 11:23 Claes Nilsson <notifications@github.com>
ha scritto:

PR #60, opened 2015-05-25
Fix missing handler

### LINKED ISSUE #55: Handler not called when proxyUri not present

dmoranj:
Fixes #55

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -34,7 +34,7 @@ function handleEnding(err) {
 
 function CoAPServer(options, listener) {
   if (!(this instanceof CoAPServer)) {
-    return new CoAPServer(options)
+    return new CoAPServer(options, listener)
   }
 
   if (typeof options === 'function') {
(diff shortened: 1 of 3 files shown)


mcollina:
Can you add a unit test for the fix, so that we cover this and we don't break it in the future?

ClaesNilsson:
I just want to state that I patched this bug fix and the bug fix in https://github.com/mcollina/node-coap/pull/61 into my local node-coap module and it works fine. That is, a node.js application acting both as a server and a client creates a coap-server that is a proxy. Peers sends requests to this proxy server and I am able to send requests back to those peers through my own local proxy. This means that the quadruple of source and destination address/port will be the same for the traffic in both directions, which makes it possible to pass firewalls.
Will https://github.com/mcollina/node-coap/pull/60 and https://github.com/mcollina/node-coap/pull/61 be merged?

PR #61, opened 2015-05-25
Fix Observes through proxy

### LINKED ISSUE #56: Proxy blocks observe responses

dmoranj:
Fixes #56

### DIFF
--- a/lib/middlewares.js
+++ b/lib/middlewares.js
@@ -12,6 +12,7 @@ var dgram           = require('dgram')
   , crypto          = require('crypto')
   , events          = require('events')
   , LRU             = require('lru-cache')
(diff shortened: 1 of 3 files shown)


lib/middlewares.js
```diff
@@ -63,6 +64,14 @@ function proxyRequest(request, next) {
   }
 }
 
+function isObserve(packet) {
+  function or(previous, current) {
+    return previous || current;
+  }
```
mcollina: can you move this or function externally?

dmoranj: Fixed in 6a69d4a

PR #65, opened 2015-07-18
Fix request error leak

neich:
I have added a 'timeout' event at RetrySend, and it is captured by the Agent that cleans up the request and avoids keeping a reference to a request that never will be completed

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -314,7 +314,10 @@ Agent.prototype.request = function request(url) {
     that._msgInFlight++
   })
 
-  req.sender.on('timeout', that.abort.bind(that, req))
+  req.sender.on('timeout', function (err) {
+    req.emit('timeout', err)
+    that.abort(req)
+  })
 
   req.sender.on('sent', function() {
     that._msgInFlight--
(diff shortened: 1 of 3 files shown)


mcollina:
This is awesome! Can you add a unit test for this?

neich:
I've been trying to add a simple test and I'm unable to pass it. Disclaimer: I've never used mocha.

I've added this test right after ```'should error after ~247 seconds'``` (```test/request.js```)

```
    it('should timeout after ~247 seconds', function(done) {
      var req = doReq()

      req.on('timeout', function(err) {
        expect(err).to.have.property('message', 'No reply in 247s')
        expect(err).to.have.property('retransmitTimeout', 247)
        done()
      })

      fastForward(1000, 247 * 1000)
    })
```

It fails. But what's worse, if at ```lib/retrySend.js``` I emit ```'error'``` and ```'timeout'``` in the reverse order (first ```'timeout'```) the my test passes ... but all the ones that listen to 'error' fail !! I'm sure I'm missing something, but I have no clue ...

neich:
Ok, I figured out. It seems that mocha checks that all emitted events have been processed.

I've added a very simple test, but maybe it should be merged with the previous one.

PR #66, opened 2015-07-19
Fix index out of range

neich:
Add guard to check if token length is > 0 before removing request reference

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -217,7 +217,7 @@ Agent.prototype._handle = function handle(msg, rsinfo, outSocket) {
   else if (block2) {
     delete that._tkToReq[req._packet.token.readUInt32BE(0)]
   }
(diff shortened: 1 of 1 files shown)


mcollina:
Can you please add a unit test?
Il giorno dom 19 lug 2015 alle 20:13 neich <notifications@github.com> ha
scritto:

mcollina:
also, would you like to be help maintaining this project? Check if you are ok with CONTRIBUTING.md, and then send a PR to add your details to the README.

neich:
After further research, I've realized that this fix is not good. The problem is the Cooja MOTE is returning a response with a token length = 0, so the right behavior would be to raise an error. You can dismiss this PR, I'll send another with a proper fix.

mcollina:
Great! I'm waiting :)

PR #68, opened 2015-07-22
Check token integrity in reponses

neich:
Added a guard to check that the token in the response is the same that the one in the request, except for 0.00 messages. Right now the behavior is to ignore messages with different tokens, so if no correct message arrives, a timeout will be emitted.


mcollina:
This PR is failing on node 0.10, because of a missing compare is not
available there. Can you please use something else, or use a polyfill?
Il giorno mer 22 lug 2015 alle 17:40 neich <notifications@github.com> ha
scritto:

neich:
Sorry, I didn't realize that the library checks againts node 0.10, I'm using 0.12. I've added a very simple polyfill to allow buffer comparison in 0.10

mcollina:
Can you please not modify Buffer's prototype, but rather use a function that calls `Buffer#compare` if it's there, or otherwise use your polyfill?

You might use https://github.com/soldair/node-buffer-compare if that is missing.

mcollina:
Maybe we should just use https://github.com/davidmarkclements/buffer-compare-polyfill and forget about it.

neich:
Ok, I've pushed my new function to compare buffers without altering Buffer.prototype, but as you say, maybe it would better to use buffer-compare-polyfill. Your call.

mcollina:
I like it this way.

PR #70, opened 2015-07-24
Fix crash when receiving duplicate CON responses

neich:
Fix crash when receiving duplicate CON responses. Right now the behavior when receiving a duplicate CON response is to ignore it and return.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -220,7 +220,7 @@ Agent.prototype._handle = function handle(msg, rsinfo, outSocket) {
       return req.response.append(packet)
     } else {
       // TODO There is a previous response but is not an ObserveStream !
-      req.response = null
+      return
     }
 
   }
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -214,9 +214,15 @@ Agent.prototype._handle = function handle(msg, rsinfo, outSocket) {
   }
 
   if (req.response) {
-    // it is an observe request
-    // and we are already streaming
-    return req.response.append(packet)
+    if (req.response.append) {
+      // it is an observe request
+      // and we are already streaming
+      return req.response.append(packet)
+    } else {
+      // TODO There is a previous response but is not an ObserveStream !
+      req.response = null
+    }
+
   }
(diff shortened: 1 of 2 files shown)


lib/agent.js
```diff
@@ -214,9 +214,15 @@ Agent.prototype._handle = function handle(msg, rsinfo, outSocket) {
...
-    // and we are already streaming
-    return req.response.append(packet)
+    if (req.response.append) {
+      // it is an observe request
+      // and we are already streaming
+      return req.response.append(packet)
+    } else {
+      // TODO There is a previous response but is not an ObserveStream !
+      return
+    }
```
mcollina: I think ignoring this packet is ok, but probably it might make sense to check in the spec.

neich: The spec says:

https://tools.ietf.org/html/rfc7252#section-4.5
```
4.5.  Message Deduplication

   A recipient might receive the same Confirmable message (as indicated
   by the Message ID and source endpoint) multiple times within the
   EXCHANGE_LIFETIME (Section 4.8.2), for example, when its
   Acknowledgement went missing or didn't reach the original sender
   before the first timeout.  The recipient SHOULD acknowledge each
   duplicate copy of a Confirmable message using the same
   Acknowledgement or Reset message but SHOULD process any request or
   response in the message only once.  This rule MAY be relaxed in case
   the Confirmable message transports a request that is idempotent (see
   Section 5.1) or can be handled in an idempotent fashion
```
We shoud process the message in the response only once, so I think it's ok to just return, right?

mcollina: I would prefer the request to be ACKed in this case, or am I missing
something?
Il giorno ven 24 lug 2015 alle 18:30 neich <notifications@github.com> ha
scritto:

neich: The CON response is already ACKed when we reach point of detecting the duplicate. If you look at Agent._handle, the first thing done is to send an ACK if the response is CON.

mcollina:
Great, thanks! :)

PR #72, opened 2015-07-27
normalize option names for proper aliasing; closes #71

### LINKED ISSUE #71: case-insensitive option matching
I have something that's trying to send option `content-type`.  This results in:

```
    Error: Unknown string to Buffer converter for option: content-type
    at module.exports.toBinary (/Volumes/alien/projects/digs/tmp/node_modules/coap/lib/option_converter.js:18:11)
    at OutMessage.setOption (/Volumes/alien/projects/digs/tmp/node_modules/coap/lib/helpers.js:46:16)

```

@mcollina Would you accept a PR to just run `.toLowercase()` across the option names?

boneskull:
- use `capitalize` module
- add test

Instead of lowercasing everything, capitalize instead.

I'm gonna go ahead and send this PR anyway, because my wager is that the names are case-insensitive.  The alias(es) won't work without normalization anyway.

see #71 and coap-technology/coap-technology.github.io#9

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -14,6 +14,7 @@ var toBinary    = require('./option_converter').toBinary
       , '0.03': 'PUT'
       , '0.04': 'DELETE'
     }
+  , capitalize  = require('capitalize')
 
 module.exports.genAck = function(request) {
   return {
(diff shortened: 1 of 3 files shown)


boneskull:
gonna ditch this for one which fixes #73 as well, since they are related sorta

PR #74, opened 2015-07-28
guard against closing closed/never-opened socket

OJFord:
Simple if-guard. Throwing an error probably not desirable, since when calling `close` we want it closed; if it already is closed then there's no error in the outcome, just an odd request.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -161,26 +161,29 @@ CoAPServer.prototype.listen = function(port, address, done) {
     address = null
   }
 
+  if (this._sock) {
+    if (done)
+      done(new Error('Already listening'))
+    else
+      throw new Error('Already listening')
+
+    return this
+  }
+
   if (address && net.isIPv6(address))
     this._options.type = 'udp6'
 
   if (!this._options.type)
     this._options.type = 'udp4'
 
-  if (this._sock && done)
-    done(new Error('Already listening'))
-  else if (this._sock)
-    throw new Error('Already listening')
-  else {
-    this._sock = dgram.createSocket(this._options.type, handleRequest(this))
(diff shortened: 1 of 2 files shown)


mcollina:
I would probably have it return if there is no socket.
Probably, we should also guard listen the same way (so the socket does not get overwritten).

Can you please add a unit test for this?

OJFord:
Please do throw out / critique the tests - never used Mocha (still new to Node) so they may well be complete rubbish!

Unsure why the seemingly unrelated test fails -- even less sure why adding tests changed it from not working only on v.1 to, fine there but not on iojs. I suppose because it's testing timing - but I don't know specifically why it fails or how to fix it.

lib/server.js
```diff
@@ -161,40 +161,40 @@ CoAPServer.prototype.listen = function(port, address, done) {
...
-    that.emit('error', error)
-  })
-
-  this._sock.bind(port, address || null, done || null)
-  this._port = port
-  this._address = address
+  if (this._sock && done)
+    done(new Error('Already listening'))
+  else if (this._sock)
+    throw new Error('Already listening')
```
mcollina: I would move this block before the address checking, because I do not want `this._options` to be modified.

lib/server.js
```diff
@@ -161,40 +161,40 @@ CoAPServer.prototype.listen = function(port, address, done) {
...
+  else {
```
mcollina: there is no need to have an else here, just do an early return from the previous block

mcollina:
The tests are good, I have left some other notes in there!

mcollina:
Thanks!! :)

Issue #73, opened 2015-07-27
ignore charset info when converting content format to binary


boneskull:
a content format `Content-Type: application/json; charset=utf8` will blow up.  I believe we can strip the semicolon (inclusive) to the end during conversion.

>    CoAP does not include a separate way to convey content-encoding
>    information with a request or response, and for that reason the
>    content-encoding is also specified for each identifier (if any).  If
>    multiple content-encodings will be used with a media type, then a
>    separate Content-Format identifier for each is to be registered.
>    Similarly, other parameters related to an Internet media type, such
>    as level, can be defined for a CoAP Content-Format entry.

boneskull:
alright, I'm not sure what to do with this.  we should allow the options to pass through verbatim.  however, if we strip the charset info from the format in order to match it against, say, `application/json`, then there's a mismatch with buffer length.  I don't understand why the pre-registered formats have the byte values they do.

mcollina:
The formats are part of the coap spec, they are represented as a binary.

Here are the definitions in this library: https://github.com/mcollina/node-coap/blob/66a84eabeb0ed04b7b7d6acec8ea0252a153fb6e/lib/option_converter.js#L73-L78

mcollina:
I don't get the mismatch for the buffer length.

boneskull:
I'll come up with a reproducible case

PR #75, opened 2015-07-30
normalize names, encoding information & writeHead; closes #71 and #73

### LINKED ISSUE #71: case-insensitive option matching

boneskull:
- Add `capitalize` package as dependency
- To normalize option names, capitalize each word of the name.  "content-format" becomes "Content-Format", etc.  special case for "ETag", which was added to the alias map.
- Fixed an undefined variable in `option_converter`'s `contentFormatToBinary()`
- Added `OutgoingMessage.prototype.writeHead()` as to provide more compatibility with the `http` module
- Encoding information can be sent in *any* `Content-Format` value, but it is discarded.  From what I can gather from the RFC7252 and other RFCs I read, we only seem to be *required* to accept the encoding information with `text/plain`; support any other format would be optional, but since sending a type such as `application/json; charset=utf-8` is so common, we may as well
- Updated `README.md` with `writeHead()` info.

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,6 +307,9 @@ for all the possible options.
 #### message.reset()
 Returns a Reset COAP Message to the sender. The RST message will appear as an empty message with code `0.00` and the
 reset flag set to `true` to the caller. This action ends the interaction with the caller.
+
+#### message.writeHead(code, headers)
+Functions somewhat like `http`'s `writeHead()` function.  If `code` is does not match the CoAP code mask of `#.##`, it is coerced into this mask.  `headers` is an object with keys being the header names, and values being the header values.   
 -------------------------------------------------------
 <a name="incoming"></a>
 ### IncomingMessage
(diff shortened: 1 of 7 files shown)


mcollina:
Tests on Travis are failing :(

boneskull:
sorry, this should fix it

boneskull:
i'm not *entirely* sure `writeHead()` is doing what it should, so there may be another forthcoming PR addressing that.

mcollina:
I think it does :).

If you are ok with CONTRIBUTING.md, I'm adding you to the Contributors list.

boneskull:
@mcollina sure, thanks

mcollina:
Send a PR to add your details in the README

Issue #6, opened 2013-10-11
Multicast Support


mcollina:
At the moment, __node-coap__ does not support multicast communication, as it it does not expose any API to add/remove the membership to a group.

edhemphill:
@mcollina Partial server-side support is here... I'll try to do a pull request as soon as I have more time. 
https://github.com/WigWagCo/node-iotivity/tree/master/lib/node-coap

mcollina:
@edhemphill I suggest you to depend on the library via NPM, and not to maintain your custom fork :). If you need to depend to your own version for some time, you can use a git dependency in package.json.

Anyway, this is an OPEN Open Source Project, so if it is ok with you I will add you to the maintainers list as soon as this is merged :).

edhemphill:
Go for it -- sorry for the delay. Yes, we would prefer to use it from NPM. Thanks!

mcollina:
Added you to this repo, feel free to contribute to a feature branch/pull request here :).

edhemphill:
Will do soon - thanks

pastukhov:
Any progress?

mcollina:
As said before, if anybody wants to contribute i will be more than happy!

neophob:
should be fixed, right?

mcollina:
Yes it is.

PR #86, opened 2015-11-11
Multicast server and request support

mast:
Here multicast request and server support is added.
Related issue https://github.com/mcollina/node-coap/issues/6

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -161,8 +161,9 @@ Agent.prototype._handle = function handle(msg, rsinfo, outSocket) {
     return
   }
 
-  if (!packet.confirmable)
+  if (!packet.confirmable && !req.multicast) {
     delete this._msgIdToReq[packet.messageId]
+  }
 
   req.sender.reset()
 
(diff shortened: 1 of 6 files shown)


mast:
It's strange that Travis fails. All tests (both new and old) pass in my environment.

neich:
Check your node version. Travis is failing with node v0.12, nut not with v0.10 ...

mast:
@neich my node is 0.10.37, so I guess that's why I don't see any fails.

mcollina:
You can use [nvm(https://github.com/creationix/nvm) to switch between multiple node versions.

mast:
It seems that newer versions of node doesn't allow to reuse same port/address for 2 sockets. It forces me to remove multicast test cases related to 2 servers started on the same host.

lib/agent.js
```diff
@@ -330,12 +342,21 @@ Agent.prototype.request = function request(url) {
...
     }
   })
 
+  req.multicastTimer = setTimeout(function() {
+    that._msgInFlight--
+    if (that._msgInFlight === 0) {
+      that._doClose()
+    }
+  }, multicastTimeout)
+
```
mcollina: This multicastTimer should be allocated only if it is a multicast request.

mast: It's very good point. Thank you for noticing!
I pushed the update.

mcollina:
Perfect!

PR #91, opened 2015-11-16
js hint fixes for agent.js

neophob:
Hi

I applied some jshint fixes so the whole code can be validated with jshint. I you like this I'll update all other classes - maybe we can easy spot other typos in the code. what do you think?

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -6,210 +6,210 @@
  * See the included LICENSE file for more details.
  */
 
-var bl              = require('bl')
-  , util            = require('util')
-  , events          = require('events')
-  , dgram           = require('dgram')
-  , parse           = require('coap-packet').parse
-  , generate        = require('coap-packet').generate
-  , URL             = require('url')
-  , IncomingMessage = require('./incoming_message')
-  , OutgoingMessage = require('./outgoing_message')
-  , ObserveStream   = require('./observe_read_stream')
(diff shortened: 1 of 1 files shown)


mcollina:
I'm in for using [standard](http://npm.im/standard), rather than jshint.
Feel free to send a PR for that if you want :).

boneskull:
Indeed I'd prefer some eslint-derived thing over jshint

boneskull:
@mcollina FWIW...

If your intention is to simply have something work, `standard` is appropriate, since you don't need to worry about configuration.

If your intention is to avoid semicolons, `standard` is also appropriate, since it enforces this.

If your intention is accessibility, I'd argue vehemently against `standard`, since from the data I've seen, the overwhelming majority of JS developers *always* use semicolons.

If you don't care about semicolons so much and just want something that works: [semistandard](https://www.npmjs.com/package/semistandard).

mcollina:
This is a democracy :). I just want something that works and has no config whatsoever.

I firmly believe that semicolons are bad, but this is OOSS, and everything is up to discussion. If most of the contributors prefers semistandard, then let's go for it. IMHO this code is really not accessible, and not because of missing semicolons.

Anybody else has an opinion? standard or semistandard?

boneskull:
:+1: for semistandard.

PR #92, opened 2015-11-16
standard updates

neophob:
Uff, I modified the code so the standard code style matches. Quite a couple if import could be removed. Most changes are quite non-invasive, but take a look at:
- middleware.js: function isObserve - wrong parameter
- helpers.js: function parseBlock2 - moreBlock2 flag
- agent.js: function _init - packet reference was unused

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,11 @@ If it is an object:
 - `proxyUri`: adds the Proxy-Uri option to the request, so if the request is sent to a
   proxy (or a server with proxy features) the request will be forwarded to the selected URI.
   The expected value is the URI of the target. E.g.: 'coap://192.168.5.13:6793'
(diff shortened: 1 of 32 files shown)


neophob:
some tests are really flacky, for example

```
1) end-to-end formats should pass the 'Content-Format: application/link-format' option to the server if passed alongside the url:
     Error: done() called multiple times
```

any hints?

mcollina:
Tests are passing on current master.
However, some of them are flaky, but they should pass, so it's probably some changes done by the reformatting.

mcollina:
BTW, do you prefer standard or semistandard?

neophob:
Yes, I'll take a look at them soon. Well about standard or semistandard - I worked with jshint in the past but standard is fine too - one of them is better than nothing an both help to detect bugs.

PR #95, opened 2015-12-02
cache block2 requests

neophob:
This PR implements packet caching for block2 requests. If a block2 transfer starts, block 0 request gets processed, block 1 request is served automatically from the server and does not need any handling in the upper layer.

nice sideeffect of this PR, I bumped mocha and now the unit tests should be fine


neophob:
there is still a nasty bug inside this code (at least on a ARM system): it looks like the device does not find the cached entry in the LRU cachen - EVEN if a getKeys() shows this key in the cache.

neophob:
It works fine now - currently only the last block2 request is cached. But the speed gain is massive, especially if you have a small packet size (64bytes, 6lowpan). Can anyone else test this?

PR #99, opened 2015-12-20
Allow more than one multicast server to bind on same port and address

paulobrizolara:
I was trying to run the 'multicast_client_server' example, but was throwing 'EADDRINUSE' bind error.
I ended founding the explanation on this thread: https://github.com/mdns-js/node-mdns-js/issues/31
Just one change on 'lib/server.js' and the example worked. So i made this little pull request.

Hope that helps. 

P.s.: running more than one server in different process also work
Note: i tested on node v0.12.7

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -180,7 +180,7 @@ CoAPServer.prototype.listen = function(port, address, done) {
   if (!this._options.type)
     this._options.type = 'udp4'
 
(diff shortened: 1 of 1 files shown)


paulobrizolara:
Looks like node v0.10.x does not support passing an object as option in 'dgram.createSocket' (https://travis-ci.org/mcollina/node-coap/jobs/98028649) :/

Maybe there is some way to check the running node version (i don't know how).

PR #102, opened 2016-01-23
add piggyback timing to parameters.js

### LINKED ISSUE #100: Force piggybacking.
I can't see a way to force piggybacked responses, which is desirable for working with motes (by reducing the amount of packets and thus the probability of loss).

Is it possible?

neophob:
should fix #100

### DIFF
--- a/test/server.js
+++ b/test/server.js
@@ -952,13 +952,63 @@ describe('server', function() {
 })
 
 describe('piggyback timing option', function() {
-  it('send message', function(done) {
-    var piggyBackTimeout = 42
 
-    var server = coap.createServer({ piggybackReplyMs: piggyBackTimeout })
+  var server
+  var port
+  var client
+  var clientPort
 
-    expect(server._options.piggybackReplyMs).to.eql(piggyBackTimeout)
+  beforeEach(function(done) {
+    port = nextPort()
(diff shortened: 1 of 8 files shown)


mcollina:
I would put this in the server options, rather than in the parameters. This should be really easy to change and customize for each server.

neophob:
make sense, ill update the pr

neophob:
done, the piggyback timeout can be defined as server option, example:
 ```coap.createServer({ piggybackReplyMs: 200 })```

test/server.js
```diff
@@ -946,7 +946,20 @@ describe('server', function() {
...
+
+describe('piggyback timing option', function() {
+  it('send message', function(done) {
+    var piggyBackTimeout = 42
+
+    var server = coap.createServer({ piggybackReplyMs: piggyBackTimeout })
+
+    expect(server._options.piggybackReplyMs).to.eql(piggyBackTimeout)
+    server.close()
+    done()
```
mcollina: It would be better if we test this with a `setTimeout`, and add a test also for the default.

neophob:
Yes, testing was done a bit in a hurry. fixed it now and improved parameter check

mcollina:
Thanks, release as 0.13.1.

PR #103, opened 2016-02-10
Fix undefined var problem in isObserve()

zacwbond:
When I tried to use the coap proxy server, I got a curious error.  I created a proxy and target servers in separate processes, based on the example proxy code, and then used the coap-cli module to construct a message to the target through the proxy (i.e., I added the `proxyUri` option).  The message and response worked, but shortly after the client received the response, the proxy and the target got locked into an endless loop of debug like this: 

`Target receives [request is not defined] in port [8976] from port [6780].`

I tracked the problem down to isObserve(), where it appears `request` is being used instead of `packet`, and which is also not even testing anything because it doesn't return a value.  

All of the unit tests pass in spite of this.  Apparently, that's because it takes too much time for the error to percolate up, and I guess none of the tests check for unexpected requests anyway.

I could use some advice on the test I added.  I had to disable the sinon fake timers to get it to pass (otherwise, it would hang forever), which is probably bad form.  I'm sure there is a way to use them and make it work, but I'm not familiar enough with them to know how.

### DIFF
--- a/lib/middlewares.js
+++ b/lib/middlewares.js
@@ -51,7 +51,7 @@ function proxyRequest(request, next) {
 }
 
 function isObserve(packet) {
-  packet.options.reduce(or, false);
+  return packet.options.reduce(or, false);
 }
 
 function handleProxyResponse(request, next) {
--- a/lib/middlewares.js
+++ b/lib/middlewares.js
@@ -51,7 +51,7 @@ function proxyRequest(request, next) {
 }
 
 function isObserve(packet) {
-  request.packet.options.reduce(or, false);
+  packet.options.reduce(or, false);
 }
 
 function handleProxyResponse(request, next) {
--- a/test/proxy.js
+++ b/test/proxy.js
@@ -112,7 +112,7 @@ describe('proxy', function() {
   it('should resend notifications in an observe connection', function(done) {
     var counter = 0,
         req;
-
+    clock.restore()
     req = sendObservation()
 
     req.on('response', function(res) {
--- a/test/proxy.js
+++ b/test/proxy.js
@@ -144,6 +144,12 @@ describe('proxy', function() {
   })
 
   it('should return the target response to the original requestor', function(done) {
+    var counter = 0;
+    clock.restore()
+    setTimeout( function() {      
+      done();
+    }, 300 );
+    
     send(generate({
       options: [{
         name: 'Proxy-Uri'
@@ -152,13 +158,14 @@ describe('proxy', function() {
     }))
 
     target.on('request', function(req, res) {
+      counter++;
+      expect(counter).to.eql(1);   
       res.end('The response')
     })
 
     client.on('message', function(msg) {
       var package = parse(msg)
       expect(package.payload.toString()).to.eql('The response')
-      done()
     })
   })
 


mcollina:
This is breaking unit tests on travis: can you check? https://travis-ci.org/mcollina/node-coap/builds/108278145

zacwbond:
The error is an unhandled `ECANCELED` in `send` that happens as a result of calling `server.close()` in `afterEach`.  It happens only after the "resend notifications in an observe connection" test.  

I don't know what exactly triggers an `ECANCELED`, but if I do a `clock.restore()` before the test, it doesn't happen.  Is this a timing thing?  I see the same error on a bunch of previous commits, but I don't see how it has been fixed in the past...

zacwbond:
Well...that didn't work.

The problem seems to be that `sendObservation` opens a new socket via the `request()` mechanism, and somehow that socket is breaking after the end of the observable test.  In particular, strace shows that there are asynchronous ACKs going on from the `observer` to the `proxy` and from the `proxy` to the `target`.  I think `ECANCELED` happens when it breaks during the ACK send, and I think the "not running" error happens when the socket breaks and then an attempt to send the ACK is made.  

I think the call to done() needs to be somehow hooked into the `finish` event of the server and the target, but my attempt to do so just doesn't work (I get `finish` after one ACK and then never see another finish again).

I give up for now.  Maybe I will look at this again one day.

dmoranj:
It seems I have been working on this exact same problem (I forgot to check PRs before starting), as you can see in PR #104. The `isObserve()` function was quite wrong, as it was making a strange or of objects. I modified it and checked with Wireshark this time to ensure the interchanged packets were the appropriate ones. 

The ECANCELLED problem remains, anyway, but not in all Node versions. I haven't been able to reproduce it in v0.10.42, while it's easily reproducible with versions v0.12.10 and above. I'll try to fix the errors in my PR considering yours.

dmoranj:
This raises another interesting question, BTW. What versions of Node.js is node-coap expected to support? 0.10 to 5?

mcollina:
This should support node v0.10 -> v5.

mcollina:
@dmoranj does your pr supersedes this?

dmoranj:
@mcollina Yes it does. I'm facing the exact same problem with the ECANCELED, and the solution for the isObserve() problem needs for some other corrections to work properly. We can continue the discussion there if you prefer.

dmoranj:
Concerning the supported versions of Node.js, I would suggest limiting to v0.10 -> v4, as, even skipping the whole proxy test suite,  there are plenty of errors runnning v5. v0.12 and v4.2.4 on the other hand, seem to run the tests, if we skip one faulty test in the proxy code. I would suggest to disable v5 from Travis and add an issue for the v5 support, as it may have some more work to do than simply fixing a couple of bugs.

Working in the proxy tests, they seem to have some interaction with tests in other suites (maybe something is left open or some value is globally changed) as, with @zacwbond solution it seems tu run fine on its own, but failed when running along the other tests. Concerning the referred solution, it really works but I'm afraid I don't understand why (as I'm not quite sure what the error is in first place). I'm sure the best solution would be to refactor the test but I'm facing the same problem with the `finish` events, so it doesn't seem clear to me how to modify that. I'll be trying to make things work on versions up to v4, and wait to hear your opinion about v5.

mcollina:
@dmoranj I don't have any bandwidth for this ATM.

dmoranj:
@mcollina Then I think it's better to temporarily disable v5 while we work on the integration. I have disabled it in the other PR, and I'm going to create an issue right now. I think I may have some time to look at this this week.

dmoranj:
I'm also closing this PR, as it completely overlaps with #103

Issue #108, opened 2016-03-10
Do not send ACK's for non-confirmable messages


neophob:
I would like to add an option to the server, that the coap server does not send ACK messages for non confirmable messages. The reason for this feature request is, that the radio layer already sends ACK's, thus the ACK's from CoAP are redundant and waste bandwidth.

I tried to implement it but somehow got stuck, when I take a look at the package structure there are two confirm value, one on top and the other one for the outgoing message. 

here is a first version: https://github.com/neophob/node-coap/commit/0a88f252f985feec63eca5fbeca05546db3a4479#diff-c945a46d13b34fcaff544d966cffcaba

I just started to add unit tests but the results do not make any sense to me, for example this test fails:

```
  it('should send ACK for confirmable message, server disabled send non confirmable packets', function(done) {
    //GIVEN
    server = coap.createServer({ sendAcksForNonConfirmablePackets: false })
    server.listen(port)
    server.on('request', function(req, res) {
      res.end()
    })

    //WHEN
    send(generate({
        confirmable: true
      , messageId: 4242
      , token: new Buffer(5)
    }))

    //THEN
    client.on('message', function(msg) {
      done()
    })
  })
```

as the generated packet looks like this:

```
{ messageId: 4242,
  token: <Buffer b0 4e 80 01 01>,
  options: [],
  confirmable: false, << ? why
  ack: true,
  reset: false,
  code: '2.05',
  payload:
   OutMessage {
  ...
     _request:
      { code: '0.01',
        confirmable: true,
```

any hints? it looks like the test code generates a wrong package...

PR #117, opened 2016-04-29
opt out ack packages for non confirmable messages

neophob:
RFC, this is an update for Issue #108 

missing a function test, but i would like to get some feedback first

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -26,6 +26,7 @@ var dgram           = require('dgram')
   , createBlock2    = require('./helpers').createBlock2
   , getOption       = require('./helpers').getOption
   , isNumeric       = require('./helpers').isNumeric
+  , isBoolean       = require('./helpers').isBoolean
   , middlewares     = require('./middlewares')
   , debug           = require('debug')('CoAP Server')
 
@@ -66,6 +67,9 @@ function CoAPServer(options, listener) {
     this._options.piggybackReplyMs = parameters.piggybackReplyMs
   }
 
+  if (!isBoolean(this._options.sendAcksForNonConfirmablePackets)) {
+    this._options.sendAcksForNonConfirmablePackets = parameters.sendAcksForNonConfirmablePackets
+  }
   this._middlewares.push(middlewares.handleServerRequest)
 
   // Multicast settings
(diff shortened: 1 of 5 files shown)


lib/server.js
```diff
@@ -66,6 +67,9 @@ function CoAPServer(options, listener) {
     this._options.piggybackReplyMs = parameters.piggybackReplyMs
   }
 
+  if (!isBoolean(this._options.sendAcksForNonConfirmablePackets)) {
+    this._options.sendAcksForNonConfirmablePackets = parameters.sendAcksForNonConfirmablePackets
+  }
```
mcollina: Not sure why you need this `isBoolean`  thing

lib/server.js
```diff
@@ -289,10 +293,13 @@ CoAPServer.prototype._handle = function(packet, rsinfo) {
...
                   packet, packet.ack || !packet.confirmable), buf)
-
     buf.sender = sender
 
-    sender.send(buf, packet.ack || packet.reset || packet.confirmable === false)
+    if (!that._options.sendAcksForNonConfirmablePackets && packet.confirmable === false){
+      debug('OMIT ACK PACKAGE')
+    } else {
+      sender.send(buf, packet.ack || packet.reset || packet.confirmable === false)
+    }
```
mcollina: rewrite this in positive logic, with the `send` case first.

mcollina:
I'm 👍 on this, let me know when you have added the other test.

lib/server.js
neophob: done, thanks for the input

lib/server.js
neophob: just to be sure:
```
  if (!isBoolean(this._options.sendAcksForNonConfirmablePackets)) {
    this._options.sendAcksForNonConfirmablePackets = parameters.sendAcksForNonConfirmablePackets
  }
```

if the server options is invalid, see test case "define invalid sendAcksForNonConfirmablePackets setting".

mcollina: I mean, why you had to define a function somewhere to check if it was a boolean

lib/server.js
mcollina: It does not seem done here, have you pushed?

mcollina:
Oh sorry for the weird comment, Github was showing me old stuff.

Anyway, :+1: for me. Can you remove the given, then, when comments in the tests?

neophob:
sure, I've removed those comments and also updated the readme file

Issue #120, opened 2016-05-24
Blockwise2 does not work with multiple simultaneous clients


jkivilin:
I've run in following problem with node-coap.

Client A starts large blockwise2 transfer from CoAP server. While client A is still transferring, client B makes small request to server. Now, block2cache entry for client A is cleared because client B request, and next blockwise2 request (block2.num >= 1) from client A goes back to upper level. Upper level generates different payload for client A and transfer gets corrupted and fails. Client A then retries large blockwise2 transfer from start. Now while client A is still transferring, client C makes small request to server... and so on.

Checking Etag will not solve issue at client as CoAP server has tens of active clients and server needs to scale to have 1000 to 10000 clients, ~10-100 new requests per second. Our large blockwise2 transfer can take several seconds, and will practically always end up failing.

neophob:
fixed

PR #121, opened 2016-05-24
Fix blockwise2 transfer for multiclient use-case; change single-entry block2Cache to multi-entry cache

jkivilin:
Proposed fix for issue #120

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -93,7 +93,7 @@ function CoAPServer(options, listener) {
   })
 
   this._series = series()
-  this._block2Cache = {key:'', payload:''};
+  this._block2Cache = {}
 
   if (listener)
     this.on('request', listener)
@@ -236,7 +236,21 @@ CoAPServer.prototype.close = function(done) {
     this._lru.reset()
   }
 
+  // cancel cache entry expiry timers
+  for (var k in this._block2Cache) {
+    if (this._block2Cache.hasOwnProperty(k)) {
+      debug('clean-up cache expiry timer, key:', k)
+      clearTimeout(this._block2Cache[k].timeoutId)
+      delete this._block2Cache[k]
+    }
+  }
+
   return this
+}
+
+function expiry(block2cache, k) {
+  debug('delete expired cache entry, key:', k)
+  delete block2cache[k]
 }
 
 CoAPServer.prototype._handle = function(packet, rsinfo) {
@@ -305,23 +319,49 @@ CoAPServer.prototype._handle = function(packet, rsinfo) {
   request.rsinfo = rsinfo
   response.statusCode = '2.05'
   response._request = request._packet
+  response._cachekey = toCacheKey(rsinfo.address, rsinfo.port, packet)
 
-  var block2cache = this._block2Cache;
+  var block2cache = this._block2Cache
   //inject this function so the response can add an entry to the cache
   response._addCacheEntry = function(key, payload) {
-    debug('add payload to cache, key:', key);
-    block2cache.key = key;
-    block2cache.payload = payload;
+    if (block2cache.hasOwnProperty(key)) {
+      debug('reuse old cache entry, key:', key)
+      clearTimeout(block2cache[key].timeoutId) // cancel old expiry timer
+      block2cache[key].payload = payload
+    } else {
+      debug('add payload to cache, key:', key)
+      block2cache[key] = {payload: payload}
+    }
+    // setup new expiry timer
+    block2cache[key].timeoutId = setTimeout(expiry, parameters.exchangeLifetime * 1000, block2cache, key)
   }
 
   //return cached value for blockwise requests
-  var cachedResponseSend = false;
+  var cachedResponseSend = false
   if (packet.token && packet.token.length > 0) {
-    debug('check if packet token is in cache');
-    if (this._block2Cache.key === packet.token.toString('hex')) {
-      debug('found cached payload');
-      response.end(this._block2Cache.payload);
-      cachedResponseSend = true;
+    // return cached value only if this request is not the first block request
+    var block2Buff = getOption(response._request.options, 'Block2')
+    var requestedBlockOption
+    if (block2Buff) {
+      requestedBlockOption = parseBlock2(block2Buff)
+    }
+    if (!requestedBlockOption) {
+      requestedBlockOption = {num: 0}
+    }
+
+    if (requestedBlockOption.num < 1) {
+      if (this._block2Cache.hasOwnProperty(response._cachekey)) {
+        debug('first block2 request, remove old entry from cache, key:', response._cachekey)
+        clearTimeout(this._block2Cache[response._cachekey].timeoutId)
+        delete this._block2Cache[response._cachekey]
+      }
+    } else {
+      debug('check if packet token is in cache, key:', response._cachekey)
+      if (this._block2Cache.hasOwnProperty(response._cachekey)) {
+        debug('found cached payload, key:', response._cachekey)
+        response.end(this._block2Cache[response._cachekey].payload)
+        cachedResponseSend = true
+      }
     }
   }
 
@@ -329,6 +369,14 @@ CoAPServer.prototype._handle = function(packet, rsinfo) {
     debug('no cached entry found, emit request to upper layer');
     this.emit('request', request, response)
   }
+}
+
+function toCacheKey(address, port, packet) {
+  if (packet.token && packet.token.length > 0) {
+    return packet.token.toString('hex') + '/' + address + ':' + port
+  }
+
+  return null
 }
 
 function toKey(address, port, packet, appendToken) {
@@ -416,7 +464,7 @@ OutMessage.prototype.end= function(payload) {
 
   // cache it
   if (this._request.token && this._request.token.length > 0) {
-    this._addCacheEntry(this._request.token.toString('hex'), payload)
+    this._addCacheEntry(this._cachekey, payload)
   }
   OutgoingMessage.prototype.end.call(this, payload.slice((requestedBlockOption.num)*requestedBlockOption.size, (requestedBlockOption.num+1)*requestedBlockOption.size))
 };
--- a/test/blockwise.js
+++ b/test/blockwise.js
@@ -6,22 +6,55 @@
  * See the included LICENSE file for more details.
  */
 
-var coap = require('../')
+var coap        = require('../')
+  , parse       = require('coap-packet').parse
+  , generate    = require('coap-packet').generate
+  , getOption   = require('../lib/helpers').getOption
+  , parseBlock2 = require('../lib/helpers').parseBlock2
+  , dgram       = require('dgram')
 
 describe('blockwise2', function() {
   var server
     , port
+    , clientPort
+    , client
+    , bufferVal
     , payload   = new Buffer(1300)
 
   beforeEach(function(done) {
+    bufferVal = 0
     port = nextPort()
     server = coap.createServer()
     server.listen(port, done)
   })
 
+  beforeEach(function(done) {
+    clientPort = nextPort()
+    client = dgram.createSocket('udp4')
+    client.bind(clientPort, done)
+  })
+
   afterEach(function() {
     server.close()
+    client.close()
   })
+
+  function send(message) {
+    client.send(message, 0, message.length, port, '127.0.0.1')
+  }
+
+  function nextBufferVal() {
+    if (bufferVal > 255)
+      bufferVal = 0
+    return bufferVal++
+  }
+
+  function fillPayloadBuffer(buffer) {
+    for (var i = 0; i < buffer.length; i++) {
+      buffer[i] = nextBufferVal()
+    }
+    return buffer
+  }
 
   it('should server not use blockwise in response when payload fit in one packet', function(done) {
     var payload   = new Buffer(100)         // default max packet is 1280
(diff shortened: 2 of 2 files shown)


mcollina:
Can you please add a unit test?

jkivilin:
I had to add hook for delayed sending of next blockwise packet, to allow simulating parallel requests from unit test. Does that look ok?

lib/agent.js
```diff
@@ -210,7 +210,17 @@ Agent.prototype._handle = function handle(msg, rsinfo, outSocket) {
...
+        req.sender.send(generate(req._packet))
+      }
+      var blockwise_send_ctl = {
+        num: block2.num+1,
+        send_next: blockwise_send_next
+      }
+      if (!req.emit('sending_next_blockwise', blockwise_send_ctl) ||
+          blockwise_send_ctl.send_next === blockwise_send_next) {
+        blockwise_send_next();
+      }
```
mcollina: I'm not understanding what this is for, can you please explain?

jkivilin:
Unit test for parallel block2 transfers needs to be able to generate parallel requests to one server. First blockwise transfer needs not complete before second blockwise transfer is started. 

To prevent client from automatically complete block2 transfer, I add 'sending_next_blockwise' part to 'lib/agent.js:handle()' to control sending of next block2 message. If 'sending_next_blockwise' hook removes function key 'send_next' from 'blockwise_send_ctl', then 'lib/agent.js:handle()' does not send next block2 message. If hook leaves 'send_next' unmodified, then next block2 message is sent normally.

Unit-test delays block2 transfers of 'req_client1' request for 50 milliseconds to allow second request with 'req_client2' to happen while first blockwise transfer from 'req_client1' has not fully completed. Hook used in unit-test for delaying block2 messages (with additional comments):
```javascript
   req_client1.on('sending_next_blockwise', function(send_ctl) {
      // local copy of function reference
      var next_blockwise = send_ctl['send_next'];

      // Remove 'send_next' from send_ctl, to block lib/agent.js:handle()
      // from sending next block2 message
      delete send_ctl['send_next'];

      // Delay sending of next blockwise message to server
      setTimeout(function() {
        next_blockwise();
      }, 50);
    })
```

mcollina:
@jkivilin thanks for the detailed explanation! I'm not comfortable in changing the agent to support a very specific test. How about we create coap packets manually, like https://github.com/mcollina/node-coap/blob/master/test/server.js#L107-L114?

jkivilin:
Thanks for the pointer to manual coap packet generation from unit-test.

I've took out agent change and made necessary changes to unit-test for manual handling of first request.

lib/server.js
```diff
@@ -305,23 +318,54 @@ CoAPServer.prototype._handle = function(packet, rsinfo) {
...
+    }
+
+    debug('add payload to cache, key:', key)
+    block2cache[key] = {'payload': payload}
+
+    // setup clean-up timer
+    block2cache[key].timeoutId = setTimeout(function(k) {
+      debug('delete expired cache entry, key:', k)
+      delete block2cache[k]
+    }, block2cache_expiry, key)
```
mcollina: can we avoid re-instantiating this every time, and reuse it? I think block2cache iis never overwritten during the like of the response.

mcollina:
Is this done?

lib/server.js
jkivilin:
```javascript
    if (block2cache.hasOwnProperty(key)) {
      debug('reuse old cache entry, key:', key)
      clearTimeout(block2cache[key].timeoutId) // cancel old expiry timer
      block2cache[key].payload = payload
    } else {
      debug('add payload to cache, key:', key)
      block2cache[key] = {'payload': payload}
    }
    // setup new expiry timer
    block2cache[key].timeoutId = setTimeout(function(k) {
      debug('delete expired cache entry, key:', k)
      delete block2cache[k]
    }, block2cache_expiry, key)
```
Would above be better?

mcollina: You can recycle the function inside `setTimeout` at global level.

jkivilin: Here's debug output for the new unit-test with parallel requests, key-payload pair is readded to cache after each block2 message:
```
CoAP_Server initialized +1ms
  CoAP_Server no cached entry found, emit request to upper layer +4ms
  CoAP_Server add payload to cache, key: +0ms 62636465/127.0.0.1:9071
  CoAP_Server check if packet token is in cache, key: +1ms ad64864f/127.0.0.1:49238
  CoAP_Server no cached entry found, emit request to upper layer +0ms
  CoAP_Server add payload to cache, key: +0ms ad64864f/127.0.0.1:49238
  CoAP_Server check if packet token is in cache, key: +1ms ad64864f/127.0.0.1:49238
  CoAP_Server found cached payload, key: +0ms ad64864f/127.0.0.1:49238
  CoAP_Server reuse old cache entry, key: +0ms ad64864f/127.0.0.1:49238
  CoAP_Server check if packet token is in cache, key: +1ms ad64864f/127.0.0.1:49238
  CoAP_Server found cached payload, key: +0ms ad64864f/127.0.0.1:49238
  CoAP_Server reuse old cache entry, key: +0ms ad64864f/127.0.0.1:49238
  CoAP_Server check if packet token is in cache, key: +50ms 62636465/127.0.0.1:9071
  CoAP_Server found cached payload, key: +0ms 62636465/127.0.0.1:9071
  CoAP_Server reuse old cache entry, key: +0ms 62636465/127.0.0.1:9071
  CoAP_Server check if packet token is in cache, key: +52ms 62636465/127.0.0.1:9071
  CoAP_Server found cached payload, key: +0ms 62636465/127.0.0.1:9071
  CoAP_Server reuse old cache entry, key: +0ms 62636465/127.0.0.1:9071
  CoAP_Server check if packet token is in cache, key: +52ms 62636465/127.0.0.1:9071
  CoAP_Server found cached payload, key: +0ms 62636465/127.0.0.1:9071
  CoAP_Server reuse old cache entry, key: +0ms 62636465/127.0.0.1:9071
    ✓ parallel requests should not clear block2 cache for each other (162ms)
  CoAP_Server clean-up cache expiry timer, key: +2ms 62636465/127.0.0.1:9071
  CoAP_Server clean-up cache expiry timer, key: +0ms ad64864f/127.0.0.1:49238
```

mcollina: I don't understand what this should mean.

mcollina: on this comment, instead of

```js
  // setup new expiry timer
    block2cache[key].timeoutId = setTimeout(function(k) {
      debug('delete expired cache entry, key:', k)
      delete block2cache[k]
    }, block2cache_expiry, key)
```

```js
  // setup new expiry timer
    block2cache[key].timeoutId = setTimeout(expire, block2cache_expiry, key)

// somewhere else in the file, at the top level
function expire (block2cache, k) {
      debug('delete expired cache entry, key:', k)
      delete block2cache[k]
} 
```

jkivilin:
It's done

mcollina:
LGTM for me.

@dmoranj @nqd @boneskull any idea?

lib/server.js
```diff
@@ -67,6 +67,10 @@ function CoAPServer(options, listener) {
     this._options.piggybackReplyMs = parameters.piggybackReplyMs
   }
 
+  if (!this._options.block2CacheExpiryMs || !isNumeric(this._options.block2CacheExpiryMs)) {
```
boneskull: I just have some general comments & questions; I don't know enough about this bit of the protocol to speak to specifics there.

If you're going to bother checking it's a number, you may want to check that it's also a *positive* number, and probably an integer (since we're talking about milliseconds).  Also, a couple questions:

- What happens if the number is `0`?  
- What is the relationship between this value and `exchangeLifetime`?  Would there be any sort of weirdness if it was set above or below some threshold?
- Why is the default value (line 71 below) what it is?

lib/server.js
```diff
@@ -236,9 +240,23 @@ CoAPServer.prototype.close = function(done) {
     this._lru.reset()
   }
 
+  // cancel cache entry expiry timers
+  for (var k in this._block2Cache) {
+    if (!this._block2Cache.hasOwnProperty(k))
+      continue
```
boneskull: opinion: I would prefer just to throw this whole thing in a block and avoid `continue` since `continue` can generally be avoided

lib/server.js
```diff
@@ -305,23 +323,50 @@ CoAPServer.prototype._handle = function(packet, rsinfo) {
   request.rsinfo = rsinfo
   response.statusCode = '2.05'
   response._request = request._packet
+  response._cachekey = toCacheKey(rsinfo.address, rsinfo.port, packet)
 
   var block2cache = this._block2Cache;
+  var block2cache_expiry = this._options.block2CacheExpiryMs;
```
boneskull: it doesn't appear we did anything w/r/t code style, but for the sake of consistency, can you please zap this semicolon and the one above?

lib/server.js
```diff
@@ -305,23 +323,50 @@ CoAPServer.prototype._handle = function(packet, rsinfo) {
...
-    debug('add payload to cache, key:', key);
-    block2cache.key = key;
-    block2cache.payload = payload;
+    if (block2cache.hasOwnProperty(key)) {
+      debug('reuse old cache entry, key:', key)
+      clearTimeout(block2cache[key].timeoutId) // cancel old expiry timer
+      block2cache[key].payload = payload
+    } else {
+      debug('add payload to cache, key:', key)
+      block2cache[key] = {'payload': payload}
```
boneskull: some inconsistencies with quoting keys of object literals

lib/server.js
```diff
@@ -305,23 +323,50 @@ CoAPServer.prototype._handle = function(packet, rsinfo) {
...
-      response.end(this._block2Cache.payload);
-      cachedResponseSend = true;
+    // return cached value only if this request is not the first block request
+    var block2Buff = getOption(response._request.options, 'Block2')
+    var requestedBlockOption
+    if (block2Buff) {
+      requestedBlockOption = parseBlock2(block2Buff)
+    }
+    if (!requestedBlockOption) {
+      requestedBlockOption = { num: 0 }
```
boneskull: likewise inconsistency w/ spaces within object literals

lib/server.js
```diff
@@ -331,6 +376,16 @@ CoAPServer.prototype._handle = function(packet, rsinfo) {
   }
 }
 
+function toCacheKey(address, port, packet) {
+  if (packet.token && packet.token.length > 0) {
+    var result = packet.token.toString('hex') + '/' + address + ':' + port
```
boneskull: can you please inline (eliminate) `result`?

test/blockwise.js
```diff
@@ -6,23 +6,56 @@
...
     , payload   = new Buffer(1300)
 
   beforeEach(function(done) {
+    bufferVal = 0
     port = nextPort()
     server = coap.createServer()
     server.listen(port, done)
   })
 
+  beforeEach(function(done) {
```
boneskull: any reason this is in a second `beforeEach` block?

test/blockwise.js
```diff
@@ -172,4 +205,95 @@ describe('blockwise2', function() {
       res.end(payload)
     })
   })
+
+  it('parallel requests should not clear block2 cache for each other', function(done) {
```
boneskull: in general, if possible, can you make one (1) assertion per test?

boneskull:
@jkivilin thanks for the PR; I made a bunch of comments

test/blockwise.js
jkivilin: Copied from 'test/server.js', probably because 'done' passed to both server.listen and client.bind.

lib/server.js
jkivilin: I took the `piggybackReplyMs` part here and add `block2CheckExpiryMs` with same levels of checks. Maybe both should be better checked for validity. I'm not sure if is beneficial to even have this configurable, but just use `exchangeLifetime` unconditionally.

If value is too short (less than round-trip time), cached payloads will expiry and be removed before next block2 message is received. If unnessary large, cached payload stays in memory for longer than it should.

`exchangeLifetime` seems good default as that way cache entry is kept for the same length of time as maximum time for ACK for confirmable message. When client experiences connection problems, client could be in retransmit loop for next block2 request and with `exchangeLifetime` cache entry is stored reasonably long time.

test/blockwise.js
jkivilin: Sure.

lib/server.js
jkivilin: Ok

lib/server.js
jkivilin: Ok

lib/server.js
jkivilin: Ok

lib/server.js
jkivilin: Ok

lib/server.js
jkivilin: I'll change this

boneskull:
@jkivilin please let us know if/when this is ready, thanks

jkivilin:
@boneskull It's ready. I made changes based on your comments:
 - Took out `block2CacheExpiryMs` and use `exchangeLifetime` for cache expiry
 - Split single unit-test to multiple
 - Coding style fixes

mcollina:
Seem ok for me. @boneskull you ok as well?

BTW, @jkivilin would you like to help maintaining this library?

boneskull:
Yes. @jkivilin thanks. Merging

Issue #122, opened 2016-05-26
Both ACK and RST sent in response to CON message


Guimier:
I am currently testing observation cancellation with node-coap by starting observation of a resource, waiting a few messages and then call `close` on the `ObserveReadStream` object.

I found that the server never stops sending me updates because CON updates sent by the server are answered twice: first with an `ACK` and second (hence ignored) with a `RST`. According to [RFC7641#3.5](https://tools.ietf.org/html/rfc7641#section-3.5), this behavior is not allowed.

mcollina:
Good catch! I do not have time to fix this myself, but I will be really happy to review and merge a PR that address this issue.

Guimier:
Closing since the fix landed in master and my test case is now working as expected.

PR #123, opened 2016-05-27
Do not send ACK for unrecognized CON message

Guimier:
For issue #122


mcollina:
this is failing everywhere but node v5+ :(

Guimier:
Yes, I was using a recent API in the test, fixed now.

mcollina:
Perfect, thanks!

PR #126, opened 2016-06-16
Add Max-Age support

Guimier:
This option is pretty useful and defined in the CoAP RFC, I think node-coap should support it.

I don’t know what is the best way to test it?


mcollina:
Can you please add a unit test for this? The best way to test is to do something similar to https://github.com/mcollina/node-coap/blob/master/test/end-to-end.js#L220-L231.

Guimier:
Thanks for the pointer.

mcollina:
LGTM, merging and releasing.

Issue #129, opened 2016-08-31
Multicast multiple interface binding


dwarburt:
When trying to use node-coap for some multicast traffic today on multiple interfaces, I found that it cannot bind to multiple interfaces for multicast. The nodejs `socket.addMembership` documentation was recently updated (https://github.com/nodejs/node/pull/7244/files) and I suppose it's now necessary to iterate the available interfaces and bind to each one explicitly, probably using `os.networkInterfaces()`.

mcollina:
Unfortunately, I do not have time to address this issue. I'll be happy to merge any contribution in this area.

dwarburt:
So, I've got a start on this but trying to figure out how to add tests for it now: https://github.com/dwarburt/node-coap/tree/multi-multi

PR #130, opened 2016-09-01
Bind to multiple interfaces for multicast

dwarburt:
Suggestions for unit testing are welcome.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -9,6 +9,7 @@
 require('./polyfill')
 
 var dgram           = require('dgram')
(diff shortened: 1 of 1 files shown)


mcollina:
@dwarburt what do you mean? how this is supposed to be used?

I'm a bit lost on why this PR is needed. Can you please sum it up? What do you want to achieve?

dwarburt:
Sorry, it addresses issue #129

mcollina:
I don't know either. Merging.

PR #131, opened 2016-09-07
Allow using existing socket

highwall:
To support one socket as both server and client.

### DIFF
--- a/README.md
+++ b/README.md
@@ -494,6 +494,8 @@ Opts is an optional object with the following optional properties:
 * `type`: `'udp4'` or `'udp6'` if we want an Agent on an IPv4 or IPv6
   UDP socket.
 
+* `socket`: use existing socket instead of creating a new one.
+
 -------------------------------------------------------
 <a name="globalAgent"></a>
 ### coap.globalAgent
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -44,19 +44,25 @@ function Agent(opts) {
   if (!opts.type)
     opts.type = 'udp4'
 
+  if (opts.socket) {
+    delete opts.port
+  }
+
   this._opts = opts
 
-  this._init()
+  this._init(opts.socket)
 }
 
 util.inherits(Agent, events.EventEmitter)
 
-Agent.prototype._init = function initSock() {
-  if (this._sock)
+Agent.prototype._init = function initSock(socket) {
+  if (this._sock) {
     return
+  }
 
   var that = this
-  this._sock = dgram.createSocket(this._opts.type, function(msg, rsinfo) {
+  this._sock = socket || dgram.createSocket(this._opts.type)
+  this._sock.on('message', function(msg, rsinfo) {
     var packet
       , message
       , outSocket
@@ -67,6 +73,11 @@ Agent.prototype._init = function initSock() {
       message = generate({ code: '5.00', payload: new Buffer('Unable to parse packet') })
       that._sock.send(message, 0, message.length,
                       rsinfo.port, rsinfo.address)
+      return
+    }
+
+    if (packet.code[0] === '0' && packet.code !== '0.00') {
+      // ignore this packet since it's not a response.
       return
     }
 
@@ -110,6 +121,9 @@ Agent.prototype._cleanUp = function cleanUp() {
 Agent.prototype._doClose = function() {
   for (var k in this._msgIdToReq)
     this._msgIdToReq[k].sender.reset()
+
+  if (this._opts.socket)
+    return
 
   this._sock.close()
   this._sock = null
--- a/lib/server.js
+++ b/lib/server.js
@@ -288,7 +288,12 @@ function expiry(block2cache, k) {
 CoAPServer.prototype._handle = function(packet, rsinfo) {
 
   if (packet.code[0] !== '0') {
-    // ignore packets that are not requests.
+    // According to RFC7252 Section 4.2 receiving a confirmable messages
+    // that can't be processed, should be rejected by ignoring it AND
+    // sending a reset. In this case confirmable response message would
+    // be silently ignored, which is not exactly as stated in the standard.
+    // However, sending a reset would interfere with a coap client which is
+    // re-using a socket (see pull-request #131).
     return
   }
 
--- a/lib/server.js
+++ b/lib/server.js
@@ -287,6 +287,11 @@ function expiry(block2cache, k) {
 
 CoAPServer.prototype._handle = function(packet, rsinfo) {
 
+  if (packet.code[0] !== '0') {
+    // ignore packets that are not requests.
+    return
+  }
+
   var sock      = this._sock
     , lru       = this._lru
     , acks      = this._acks
--- a/test/request.js
+++ b/test/request.js
@@ -81,7 +81,7 @@ describe('request', function() {
     var req = request('coap://aaa.eee:' + 1234)
 
     req.once('error', function () {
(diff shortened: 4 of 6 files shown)


mcollina:
Can you please add unit tests for this?

mcollina:
Great work!!

README.md
```diff
@@ -494,6 +494,8 @@ Opts is an optional object with the following optional properties:
 * `type`: `'udp4'` or `'udp6'` if we want an Agent on an IPv4 or IPv6
   UDP socket.
 
+* `sock`: use existing sock instead of creating a new one.
+
```
mcollina: Can you please use socket instead of sock?

lib/agent.js
```diff
@@ -111,6 +117,8 @@ Agent.prototype._doClose = function() {
   for (var k in this._msgIdToReq)
     this._msgIdToReq[k].sender.reset()
 
+  if (this._keep) return;
```
mcollina: Can you please remove this._keep, and move the check on opts.sock in this function?

lib/server.js
```diff
@@ -255,6 +255,9 @@ function expiry(block2cache, k) {
 
 CoAPServer.prototype._handle = function(packet, rsinfo) {
 
+  // Make sure this is a request
+  if (packet.code[0] != '0') return;
+
```
mcollina: why this?

README.md
slowdone: OK, I will update.

lib/server.js
slowdone: If the code[0] is not 0, this packet is a response of a previous request from the same socket.

mcollina: Ok. Could you add a bigger comment explaining this is needed to reuse the socket?

The problem with this change is when the socket is not reused, as the spec says we need to reply with something (couldn't remember now, please check), if we couldn't recognize the message.

highwall: OK, I will do this later, for I'm really busy recently.

mcollina:
Any update on this?

radiumfl:
Dear Highwall and Mcollina, it's a wonderful feature. How to use it?
Thanks in advance! Radiumfl

neophob:
+1

highwall:
All test cases pass.

lib/agent.js
```diff
@@ -44,19 +44,24 @@ function Agent(opts) {
   if (!opts.type)
     opts.type = 'udp4'
 
+  if (opts.socket) {
+    delete opts.port;
```
GiedriusM: Project code style [CS] does not use ";" as statement separator, so it should be removed

lib/agent.js
```diff
@@ -44,19 +44,24 @@ function Agent(opts) {
...
 util.inherits(Agent, events.EventEmitter)
 
-Agent.prototype._init = function initSock() {
+Agent.prototype._init = function initSock(socket) {
   if (this._sock)
     return
 
   var that = this
-  this._sock = dgram.createSocket(this._opts.type, function(msg, rsinfo) {
+  this._sock = socket || dgram.createSocket(this._opts.type);
```
GiedriusM: [CS] Remove ";"

lib/agent.js
```diff
@@ -70,6 +75,11 @@ Agent.prototype._init = function initSock() {
       return
     }
 
+    if (packet.code[0] === '0' && packet.code !== '0.00') {
+      // ignore this packet since it's not a response.
+      return;
```
GiedriusM: CS: Remove ";"

lib/agent.js
```diff
@@ -111,6 +121,8 @@ Agent.prototype._doClose = function() {
   for (var k in this._msgIdToReq)
     this._msgIdToReq[k].sender.reset()
 
+  if (this._opts.socket) return;
```
GiedriusM: [CS] remove ";"
[CS] the return statement should be placed into a separate line

lib/server.js
```diff
@@ -287,6 +287,12 @@ function expiry(block2cache, k) {
 
 CoAPServer.prototype._handle = function(packet, rsinfo) {
 
+  // Make sure this is a request
+  if (packet.code[0] !== '0') {
+    // just ignore this packet.
+    return;
```
GiedriusM: [CS] remove ";"

test/common.js
```diff
@@ -12,4 +12,3 @@ var portCounter = 9042
 global.nextPort = function() {
   return ++portCounter
 }
-
```
GiedriusM: This empty new line fix is not related to the issue, therefore it should go into a separate pull-request and reverted in this one.

test/share-socket.js
```diff
@@ -0,0 +1,423 @@
...
+    , port
+
+  beforeEach(function(done) {
+    port = nextPort()
+    server = coap.createServer()
+    server.listen(port, function() {
+      coap.globalAgent = new coap.Agent({
+        socket: server._sock
+      });
+      done();
```
GiedriusM: [CS] This and all the other ";" in this file should be removed

highwall:
@GiedriusM modified

lib/server.js
```diff
@@ -287,6 +287,11 @@ function expiry(block2cache, k) {
 
 CoAPServer.prototype._handle = function(packet, rsinfo) {
 
+  if (packet.code[0] !== '0') {
```
GiedriusM: This section needs an explanation. Could you please add the following comment:
According to RFC7252 Section 4.2 receiving a confirmable messages that can't be processed, should be rejected by ignoring it AND sending a reset. In this case confirmable response message would be silently ignored, which is not exactly as stated in the standard. However, sending a reset would interfere with a coap client which is re-using a socket (see pull-request #131).

test/request.js
```diff
@@ -81,6 +81,7 @@ describe('request', function() {
     var req = request('coap://aaa.eee:' + 1234)
 
     req.once('error', function () {
+      coap.globalAgent.abort(req);
```
GiedriusM: [CS] remove ";"

test/share-socket.js
```diff
@@ -0,0 +1,423 @@
...
+      res.end()
+    })
+  })
+
+  describe('formats', function() {
+    var formats = [ 'text/plain', 'application/link-format',
+      'application/xml', 'application/octet-stream',
+      'application/exi', 'application/json', 'application/cbor' ]
+
+    ;['Accept', 'Content-Format'].forEach(function(option) {
```
GiedriusM: [CS] remove ";"

test/share-socket.js
```diff
@@ -0,0 +1,423 @@
...
+      , completed = 2
+
+    server.on('request', function(req, res) {
+      res.write('hello')
+      setTimeout(function() {
+        res.end('world')
+      }, 10)
+    })
+
+    ;[req1, req2].forEach(function(req) {
```
GiedriusM: [CS] remove ";"

test/share-socket.js
highwall: Removing this `;` cause error. 
```Caught exception: TypeError: Cannot read property 'forEach' of undefined```
I think that's why test/end-to-end.js also keeps this `;`

test/share-socket.js
highwall: save as above

test/share-socket.js
GiedriusM: Ah sorry, I was too hasty to ctrl+f ; and comment, without looking at it closer :)

PR #132, opened 2016-09-30
Accept EAI_AGAIN as DNS failure

guimier:
The error code apparently depends on some configuration. Adding this
one in order to have tests passing in my development environment.


mcollina:
What is your configuration? Can you please add a comment?

guimier:
I am not sure where it comes from. I can say for sure it is not related to using Node 6 (it fails with Node 4 too). Might be related to my enterprise’s DNS since I get EAI_AGAIN on two systems.

Issue #127, opened 2016-06-20
URIs are sent unnormalized


guimier:
According to the CoAP RFC, the `Uri-Path` option is a “string”, which is supposed to be encoded in the Net-Unicode (UTF-8 + NFC) format. Node-coap currently send URIs unchanged, allowing users to pass unnormalized strings.

I think `Uri-Path` (and potentially other “string” options) fall in the “imported from sources outside the CoAP protocol” category mentioned in the RFC.

neophob:
@guimier any pending work here or should I close?

PR #133, opened 2016-09-30
Normalize URI-Path components

guimier:
Fix for #127 (more work might be required for others options).


test/end-to-end.js
```diff
@@ -134,6 +134,25 @@ describe('end-to-end', function() {
...
+    server.on('request', function(req, res) {
+      try {
+        expect(req.url).to.equal('/\u210e/\u00e9')
+        done()
+        res.end()
+      } catch (e) {
+        done(e)
+      }
+    })
+  })
```
guimier: It would be better to test the raw buffer than the parsed packet, but I did not see any test doing so and am not sure which form it would have.

lib/agent.js
```diff
@@ -6,6 +6,10 @@
  * See the included LICENSE file for more details.
  */
 
+// Polyfill for `String.prototype.normalize`.
+// To be removed when Node 0.x is no longer supported.
+require('unorm')
```
guimier: Only Node 0.10 is affected; since it is almost EOL, maybe support could just be dropped for the next release.

mcollina: Can we use a ponyfill instead of changing a global?

test/end-to-end.js
mcollina: there is no need to try-catch.

guimier: Forgot to remove it completely (used it to fix the test at some point). Will be out in next commit.

lib/agent.js
guimier: I did not find a package only exporting the normalization function.

Could use `const nfc = require('unorm').nfc` instead but `String.prototype` will still be extended.

mcollina: Do we really need this polyfill? It seems `normalize` is on string since v0.12. I'm 👍 in removing support for v0.10 in the next release.

guimier: Done.

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: node_js
 sudo: false
 node_js:
-  - "0.10"
   - "0.12"
   - "4"
   - "5"
```
mcollina: Can you please remove node v5 and add node v6?

mcollina:
@guimier do you want to help maintaining this module?

Issue #135, opened 2016-11-22
running node coap server leaks on receipt of coap request


username-II:
'use strict';
var coap =require('coap');

// actual server code
var localServer =coap.createServer(undefined,function(i,o){
  o.on('error', function(err){
    console.log('coap server error: '+err.toString());
  });
  o.on('timeout', function(){
    console.log('coap server reply timeout.');
  });
  console.log('coap server uri: '+i.url);
  console.log('memory: '+JSON.stringify(process.memoryUsage()));
  return o.end();
});

// create and decode client requests
function sendQuery(func,machine,port,path,args,body,done){
  var req =coap.request({ 
    method:func, 
    hostname:machine, 
    port:port, observe:false, 
    pathname:path, query:args
  });
  req.on('error', function(err){
    console.log('coap query error: '+func+' '+machine+':'+port+path+'&'+args+' >'+err.toString());
    done(undefined);
  });
  req.on('timeout', function(){
    // error callback above is called first
    console.log('coap query timeout: '+func+' '+machine+':'+port+path+'&'+args);
  });
  req.on('response', function(res){
    console.log('coap responce to ['+func+' '+machine+':'+port+path+'&'+args+']: ' +res.code);
    var obj =null; //cborToObj(res.payload);
    done(obj);
  }.bind(this));
  // done
  req.end();
  console.log('coap query sent: '+func+' '+machine+':'+port+path+'&'+args);
};

// the coap callback
var cb =function(o){console.log('body:' +JSON.stringify(o))};

// uses arg2 as <server port>
var port =parseInt(process.argv[3], 10);
if(port <0 || port >65535 || typeof port != 'number' || isNaN(port)) port =5683;

// if arg1 is 'server' start a server
if( process.argv[2] =='server' ){
  console.log('create server port '+port);
  localServer.listen(port);
} else {
  // test spam the server
  var postsPerSec =100;
  setInterval(function(){
    // tried with PUT, POST, GET etc...
    sendQuery('POST','127.0.0.1',port,'/well-known/core','',undefined,cb);
  }, 1000/ postsPerSec);
}

// run: node minServeLeak.js server
// watch the above process memory use
// run: node minServeLeak.js
// watch server heap total and heap used memory rise and never fall...

// tried with node v4.4.7 lts & v6.3.9 lts 32b & 64b on win10 x64 & win8.1 x64

// attempted node args to force gc but leak still appears:
[...]

PR #136, opened 2016-11-26
Fix LRU cache bloat

richardms:
See #135.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -83,11 +83,10 @@ function CoAPServer(options, listener) {
   // 32 MB / 1280 = 26214
   // The max lifetime is roughly 200s per packet.
   // Which gave us 131 packets/second guarantee
-  this._lruMaxAge = (parameters.exchangeLifetime * 1000)
   this._lru = LRU({
       max: options.cacheSize || (32768 * 1024)
     , length: function(n) { return n.buffer.byteLength }
-    , maxAge: this._lruMaxAge
+    , maxAge: (parameters.exchangeLifetime * 1000)
     , dispose:  function(key, value) {
                   if (value.sender)
                     value.sender.reset()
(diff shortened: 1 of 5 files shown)


lib/server.js
```diff
@@ -323,10 +334,17 @@ CoAPServer.prototype._handle = function(packet, rsinfo) {
...
     buf.sender = sender
 
+    var key = toKey(rsinfo.address, rsinfo.port,
+        packet, packet.ack || !packet.confirmable)
+    lru.set(key, buf)
+
+    // Start timer to actively prune from LRU cache after expiration
+    buf.lru_timer = setTimeout(function () {
+      lru.get(key);
+    }, that._lruMaxAge + 1000)
```
mcollina: you can create an interval and call `unref()`, then you won't need to reset the timer every time.

lib/server.js
```diff
@@ -83,10 +83,11 @@ function CoAPServer(options, listener) {
   // 32 MB / 1280 = 26214
   // The max lifetime is roughly 200s per packet.
   // Which gave us 131 packets/second guarantee
+  this._lruMaxAge = (parameters.exchangeLifetime * 1000)
   this._lru = LRU({
       max: options.cacheSize || (32768 * 1024)
-    , length: function(n) { return n.length }
-    , maxAge: parameters.exchangeLifetime
+    , length: function(n) { return n.buffer.byteLength }
+    , maxAge: this._lruMaxAge
```
mcollina: there is no need to make `_lruMaxAge` a property, assign it directly here.

mcollina:
Also, do you have some ideas into adding a unit test for this?

lib/server.js
richardms: OK. I'll switch to an Interval called once every maxAge that runs a prune() across the whole lru.

lib/server.js
```diff
@@ -237,6 +237,13 @@ CoAPServer.prototype.listen = function(port, address, done) {
   this._port = port
   this._address = address
 
+  if (parameters.pruneTimerPeriod) {
+    // Start LRU pruning timer
+    this._lru.pruneTimer = setInterval(function () {
+      that._lru.prune()
+    }, parameters.pruneTimerPeriod*1000)
```
mcollina: can you please `unref()`  this?

richardms:
Please ignore the PR for the time being - I'm having trouble getting the tests to run reliably on both Windows & Linux

richardms:
OK. Working well on Linux, although Travis CI is failing for node 0.12. Do you think that's something I've done or due to a changing dependency?

mcollina:
I'm ok in dropping support for node v0.12.

Would you like to help maintaining/developing this module?

richardms:
Sorry to disappear - I went off sick for the second half of last week, and I've been catching up ever since.

Time is short for me at the moment, but I'm happy to help the maintenance effort when I can.

lib/server.js
mcollina: this is still not `unref()`.

richardms: it is below - one of the mock timer libraries doesn't support `unref()` so I had to put a guard around it.

PR #137, opened 2017-01-16
Do not retry non-confirmable requests

GiedriusM:
I could not find an exact statement whether non-confirmable request CAN/SHOULD/MUST [NOT] be retried. But IMO it SHOULD be sent only once, especially for multicast requests. Also, tools like Californium send only one non-confirmable request, even if retry option is selected.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -316,7 +316,7 @@ Agent.prototype.request = function request(url) {
     that._msgIdToReq[packet.messageId] = req
     that._tkToReq[that._lastToken] = req
 
-    req.sender.send(buf)
+    req.sender.send(buf, !packet.confirmable)
   })
 
   req.sender = new RetrySend(this._sock, url.port, url.hostname || url.host)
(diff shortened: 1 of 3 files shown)


mcollina:
Can you add a unit test for this?

GiedriusM:
I could, once I get back from vacation, but currently I'm tight on other tasks. If you wish this to be changed sooner, feel free to take over.
Anyway, it looks like current non-confirmable request [tests] follow "different" logic. IMO, non-cofirmable reqest should be sent once + emit('sent'), wait for response and emit('response') or timeout with emit('timeout'). What do you think?

mcollina:
Yes, I agree with you. There is no urgency, I am traveling myself.

GiedriusM:
At this moment I have a split opinion about error event after timeout for non-confirmable request. In a broad "error" sense, timeout is an error and, also, it would be the same as for confirmable requests. On the other hand, non-confirmable requests are to be expected to get lost in a normal way of operation. At first I went with the former, but then switched to the latter (basically, commit 08c4a70). The latter is the current version and could be merged, the former needs a revert + fixes to some non-confirmable req tests to handle that error event. Do you have any thoughts on this?
Also, test "should stop retrying if it receives a message" is now somewhat redundant and could be removed.

mcollina:
@GiedriusM I trust you. I have little time for maintaining this library, would you like to take over as a lead maintainer? I assume you are using this for an application.

GiedriusM:
Yes, I am using this for an application. I would gladly help fix issues that I come across, however I doubt that I would be able to contribute in a continuous development cycle.

PR #142, opened 2017-05-04
add new option for retrySend

neophob:
Currently node-coap does not allows to set separate options for client and server. My use case: coap server should retransmit 3 times, but the other way around I want only one retransmit.

Thats why this PR, it allows to set a custom retry count for a new CoAP request.

Tests are missing, I just want to get some feedback first


GiedriusM:
Looks good to me. Could you also add the tests?

neophob:
yes - will do the tests soon, sorry for the delay

neophob:
any feedback? will merge when I receive a +1

PR #150, opened 2017-07-22
Only set ETag for Block2 response if ETag has not been set manually

crux153:
Currently node-coap automatically sets ETag option for Block2 response and this overrides any manually set ETag value.

I've added a check to see if there's any ETag value set manually before overwriting them.


neophob:
@crux153 any chance for a test so we can merge this?

PR #155, opened 2017-11-11
enable strict mode, rename reserved word

neophob:
found the ìnterface` issue when doing some webpack magic

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: node_js
-sudo: false
 node_js:
   - "4"
   - "6"
   - "7"
+  - "8"
+  - "9"
--- a/index.js
+++ b/index.js
@@ -1,3 +1,4 @@
+'use strict';
 /*
  * Copyright (c) 2013-2015 node-coap contributors.
  *
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -1,3 +1,5 @@
+'use strict';
+
 /*
  * Copyright (c) 2013-2015 node-coap contributors.
  *
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -1,3 +1,5 @@
+'use strict';
+
 /*
  * Copyright (c) 2013-2015 node-coap contributors.
  *
--- a/lib/incoming_message.js
+++ b/lib/incoming_message.js
@@ -1,3 +1,5 @@
+'use strict';
+
 /*
  * Copyright (c) 2013-2015 node-coap contributors.
  *
--- a/lib/middlewares.js
+++ b/lib/middlewares.js
@@ -1,3 +1,5 @@
+'use strict';
+
 /*
  * Copyright (c) 2013-2015 node-coap contributors.
  *
--- a/lib/observe_read_stream.js
+++ b/lib/observe_read_stream.js
@@ -1,3 +1,5 @@
+'use strict';
+
 /*
  * Copyright (c) 2013-2015 node-coap contributors.
  *
--- a/lib/observe_write_stream.js
+++ b/lib/observe_write_stream.js
@@ -1,3 +1,5 @@
+'use strict';
+
 /*
  * Copyright (c) 2013-2015 node-coap contributors.
  *
--- a/lib/option_converter.js
+++ b/lib/option_converter.js
@@ -1,3 +1,5 @@
+'use strict';
+
 /*
  * Copyright (c) 2013-2015 node-coap contributors.
  *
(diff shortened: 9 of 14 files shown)


mcollina:
Tests are failing on CI.
While you are at it, would you mind adding Node 8 and 9 as well?

neophob:
Yes saw that, my wild guess is, that either the tests are flaky or the travis machine do no support ipv6. but i'll investigate and add node 8 + 9 to the ci

GiedriusM:
Isn't module code always in strict mode?

neophob:
No - after I added 'use strict' to the server.js file and ran the tests, nodejs complained about

```
# npm t

> coap@0.21.0 test /Users/michaelvogt/_code/github/node-coap
> mocha --bail --reporter spec 2>&1

/Users/michaelvogt/_code/github/node-coap/lib/server.js:232
        allAddresses(that._options.type).forEach(function(interface) {
                                                          ^^^^^^^^^
SyntaxError: Unexpected strict mode reserved word
    at Object.exports.runInThisContext (vm.js:76:16)
    at Module._compile (module.js:542:28)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/michaelvogt/_code/github/node-coap/index.js:11:23)
```

GiedriusM:
Huh, I always assumed it is :) http://www.ecma-international.org/ecma-262/8.0/index.html#sec-strict-mode-code seems to state that.

neophob:
argh about this ipv6 issue, is it possible travis changed the server env and ipv6 is a bit a second class citizen. should I disable ipv6 tests on ci or change the tests that they run on a container and use ipv6 loopback (see https://github.com/njh/travis-ipv6-test).
Or maybe we should use nock and mock out all the network requests?

neophob:
any input how to proceed with the ipv6 test issues?

- use nock
- disable
- use other vm and use loopback interface

GiedriusM:
> any input how to proceed with the ipv6 test issues?

I would go against disabling the tests, after that VM looks like the path of least resistance.

> but what does "Module code is always strict mode code" mean anyway? what is module code?

My guess the code that's processed via `require` or at least put into `module.exports`. But I didn't write the spec, so yeah...

GiedriusM:
Did a quick test and only the OSX IPv6 tests passed - https://github.com/GiedriusM/node-coap/commits/pr-fix-travis

Should we switch testing to OSX until IPv6 is properly fixed on other containers?

Ruby github is also WAR'ing this issue, although it seems that Trusty VM works for them - https://github.com/ruby/ruby/commit/4d606546902fcc32b44ea7e07461be9cf05df909

neophob:
@GiedriusM switching to OSX build is fine for me - especially compared to now where the tests fail. thanks for your help!

GiedriusM:
@neophob, IMO it's best to do travis fix in a separate PR, as it affects more than just this issue.

neophob:
make sense, so you do a sepearte PR, ill merge that

.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: node_js
-sudo: false
```
GiedriusM: Modifications to .travis.yml should probably be reverted due to #157

neophob:
Some tests are flaky:

Node 8:
`1) share-socket "after each" hook:
     Uncaught Error [ERR_SOCKET_DGRAM_NOT_RUNNING]: Not running`

Node 8:
`1) IPv6 server should receive a CoAP message specifying the type:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. `

Not sure if this PR is the cause - wait until job from @GiedriusM finishes

GiedriusM:
Yeah, they seem to be failing on #156 too :/ Wonder if it's the OSX issue or not?!

GiedriusM:
https://travis-ci.org/mcollina/node-coap/builds/304441770 Errored - maybe Travis is having internal problems?

neophob:
yes i triggered the master build. lets wait a day a see if the problem is resolved magically or not..

neophob:
OSX build failing randomly... @GiedriusM using nock might be an option, what do you think?

GiedriusM:
Yeah, I saw that... Wanted to do more tests to see if anything is doable about that, but don't have that much spare time atm. Anyway, I have no experience with nock, so can't go either for, or against it.

neophob:
its green! thanks @GiedriusM for the travis update

GiedriusM:
Let's hope it will work longer this time :)

Issue #154, opened 2017-10-19
client reset  observe failed


ghostyu:
in client side, i want to send RST packet to server.
`req.on('response', function(res) {
  console.log("---------------------response---------------------");
  console.log(res.headers);

  var rcvCount = 5;
  res.on('data', function(data) {
    console.log(data);
    rcvCount--
    if(rcvCount == 0){
      //it will throw new Error: Empty messages must be empty
      req.reset();
    }
  });
})
req.end()`

when i call req.reset(), it will throw Empty messages must be empty error

Pluuuto:
I have almost the same problem. When I am using "res.reset()" at server-end.
```
const coap = require('coap')

const server = coap.createServer()

server.on('request', (req, res) => {
  console.log(`Receive request: ${req.url}`)
  res.reset()
})

server.listen(5683, '127.0.0.1')

```

It will cause client-end throwing "Empty messages must be empty error". Using wireshark to capture and I find server-end sending NON msg with '5.00 Internal Server Error' and body 'Empty messages must be empty' after send RST msg. 

![image](https://user-images.githubusercontent.com/9493652/32530723-61d3dd76-c405-11e7-8f81-f2b8790da683.png)

NON msg has the same token with the NON request which client-end sent. I think this cause client-end throws error. I don't know why coap source code deal with reset on this way, maybe it's a bug? (reference to 'fastseries' lib)

Pluuuto:
I found this is a bug because the RST msg has token. RFC7252-4.1 specific as follows:

![image](https://user-images.githubusercontent.com/9493652/32531104-a610128c-c407-11e7-8430-dad554598d56.png)

The Error "Empty messages must be empty" is thrown by 'coap-packet' lib 'index.js':

![image](https://user-images.githubusercontent.com/9493652/32531257-5f9894c2-c408-11e7-9c68-b5fe12ae6d74.png)

And I check coap source code 'outgoing_message.js' and I find this:

![image](https://user-images.githubusercontent.com/9493652/32531284-896637aa-c408-11e7-8347-b4b33d1bc8d0.png)

**packet.token is NOT set to null**! I think this is the essential problem. After I add "packet.token = ''" at the following, the client-end doesn't thow the Error.

![image](https://user-images.githubusercontent.com/9493652/32531374-09440c9a-c409-11e7-9f3c-68f3bf13cdd7.png)

GiedriusM:
[...]

PR #156, opened 2017-11-19
Add RST packet processing fixes

GiedriusM:
This PR fixes several issues when processing RST packets, including original issue #154:
* Server sends RST packet with token
* Server sends NON '5.00' "Empty message must be empty" after the RST
* Server sends ACK after the RST
* In rare occasions client may send a RST message in response to RST
* Request does not finish after receiving RST (cleanUp is not called, so the program just "hangs" without finishing)

Could anyone review the changes? @mcollina @neophob?

Also, I was unsure how should I handle a RST as a response - at first I did a separate req.on('reset') event, but then just slapped RST into response as it does not alter the API. Which do you guys think is better?

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -70,9 +70,6 @@ Agent.prototype._init = function initSock(socket) {
     try {
       packet = parse(msg)
     } catch(err) {
-      message = generate({ code: '5.00', payload: new Buffer('Unable to parse packet') })
-      that._sock.send(message, 0, message.length,
-                      rsinfo.port, rsinfo.address)
       return
     }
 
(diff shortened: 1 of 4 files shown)


neophob:
Hi @GiedriusM - looks nice. I will test this next week to see if there any side effects.

GiedriusM:
Thanks, I'll add you as a reviewer then.

GiedriusM:
Merged with the (hopefully) working fix. Tests seem to pass without any disturbances.

neophob:
finally - could test with my app, LGTM!

GiedriusM:
Thanks, @neophob! I'll merge it once you validate the review.

PR #158, opened 2017-11-21
Fix without timeout error when using existing socket

PeterEB:
I use an Agent with existing socket. It will call `agent._cleanUp()` to set `agent._closing` to be `true` as soon as it receive a response. Then `agent._doClose()` will be called and `req.sender._timer` will be cleared when I use the same Agent to send requests. Therefore, timeout error event will never be fired. I recognize that `agent._closing` should not be set to `true` when Agent using existing socket.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -110,7 +110,8 @@ Agent.prototype._cleanUp = function cleanUp() {
   if (--this._requests !== 0)
     return
 
-  this._closing = true
+  if (!this._opts.socket)
+    this._closing = true
 
   if (this._msgInFlight !== 0)
     return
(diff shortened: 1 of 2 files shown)


neophob:
thanks for you PR @PeterEB. can you add a unit test for that case?

PeterEB:
I've added a very simple test.

GiedriusM:
Hi, @PeterEB! There has been some issues with Travis builds. Could you merge this branch with master branch, to see if the tests finally pass?

PeterEB:
I've merged master branch. The tests are pass!

neophob:
@GiedriusM should we merge?

PR #164, opened 2018-01-10
Allow url.observe to be a number

pekkanikander:
RFC 7641 Section 2 defined that the Observe option may have an integer value:
   0 (register) adds the entry to the list, if not present;
   1 (deregister) removes the entry from the list, if present.

This commit allows the url.observe to be also a JS number object,
instead of just `true`, thereby adding support for observe deregistration.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -383,7 +383,9 @@ Agent.prototype.request = function request(url) {
     }, multicastTimeout)
   }
 
-  if (url.observe)
+  if (typeof (url.observe) === "number")
+    req.setOption('Observe', url.observe)
+  else if (url.observe)
     req.setOption('Observe', null)
   else
     req.on('response', this._cleanUp.bind(this))


GiedriusM:
Thanks, @pekkanikander! Could you also add a unit test, which checks that correct values are being sent?

pekkanikander:
@GiedriusM I'm afraid I don't understand how the test framework here works.  I'm a relative beginner with JavaScript, even though I have a long programming background.  Furthermore, I will be now super busy for the next three weeks, mostly travelling.  

That said, with some help I could probably make a unit test somewhere in mid February.  If you want me to do that, please provide me with pointers to the appropriate documentation how to write these test cases, and preferably an existing test case that I could use as a starting point.

neophob:
@pekkanikander hey its mid February ;)
just take a look at the `test/agent.js` file, just add a new test that validates you change. there's no magic involved there. make sure to validate the result - the other tests should help you

sjlongland:
I've just had a look at this pull request… I too have a need to be able to send such requests, as waiting for the server to just time out or send a notification for the client to `RST` leaves resources tied up on constrained devices for too long.

https://github.com/widesky/node-coap/commits/feature/WSHUB-357-observe-deregister is my attempt at addressing this.

This should also help with issue #195 as well: the `.close()` method of `ObserveReadStream` has a parameter, `eager` added.  If truthy, the `ObserveReadStream` emits a `deregister` event which the `Agent` class picks up.  `Agent` then submits a new request to the CoAP server de-registering the client.

I'll have to do some more testing on this, but hopefully this is along the right lines to address both issues.

Apollon77:
@pekkanikander Are you ok will closing this PR because your change is bacially also in #214 but including tests and such

pekkanikander:
> @pekkanikander Are you ok will closing this PR because your change is bacially also in #214 but including tests and such

Ok for me

PR #177, opened 2018-04-28
Inherit ObserveWriteStream from OutgoingMessage as per API documentation

Merisho:
As [documentation for ObserveWriteStream](https://github.com/mcollina/node-coap#observewrite) says it implements the Writable Stream and OutgoingMessage interfaces, but in code there hasn't been any OutgoingMessage interface implementation (or inheritance in our case).
The issue created problem with `piggybackReplyMs `- this config property wasn't used for sending empty ACK in case of separate response ([RFC](https://tools.ietf.org/html/rfc7252#section-5.2.2)) so we needed to manually send piggybacked response to acknowledge client and then send second response with actually payload and client had to handle 2 responses every time

### DIFF
--- a/lib/observe_write_stream.js
+++ b/lib/observe_write_stream.js
@@ -8,12 +8,14 @@
  * See the included LICENSE file for more details.
  */
 
-var Writable = require('readable-stream').Writable
-  , util       = require('util')
-  , helpers    = require('./helpers')
+var Writable        = require('readable-stream').Writable
+  , util            = require('util')
+  , OutgoingMessage = require('./outgoing_message')
+  , helpers         = require('./helpers')
 
 function ObserveWriteStream(request, send) {
   Writable.call(this)
+  OutgoingMessage.call(this, request, send)
 
   this._packet = {
       token: request.token
@@ -39,6 +41,7 @@ function ObserveWriteStream(request, send) {
 }
 
 util.inherits(ObserveWriteStream, Writable)
+util.inherits(ObserveWriteStream, OutgoingMessage)
 helpers.addSetOption(ObserveWriteStream)
 
 ObserveWriteStream.prototype._write = function write(data, encoding, done) {


Merisho:
Tests on Travis are failing, however tests on my local machine are OK, need to investigate

neophob:
@Merisho would you merge with latest master to check if the tests are green?

Merisho:
@neophob Rebased on master with my latest test suite fix, Travis tests are OK

Merisho:
@neophob Will you merge this PR? In other case please advise whom I can contact to request review and merge.
Thank you very much

neophob:
@GiedriusM might be able to help. but a unit test to show what changed make's it easier to understand why this PR is needed

Apollon77:
@Merisho Yes, please provide more information on the benefits of this change ... or it only "to have it in"?

JKRhb:
@Apollon77 Hmm, do you think we should close this PR? In this case the readme should probably be updated and the reference to `OutgoingMessage` should be removed.

Apollon77:
In fact yes ... I still do not really understood the benefit if it was working before -> then readme should be updated. Or we include it to include it ... but I do nt know the effects of it ...

JKRhb:
> 
> 
> In fact yes ... I still do not really understood the benefit if it was working before -> then readme should be updated. Or we include it to include it ... but I do nt know the effects of it ...

Yeah, better be sure not to break anything. Adjusting the readme to the implementation is probably safer than doing it the other way around ;)

I would then provide a PR for the readme changes and after that PR is merged this one can be closed.

Apollon77:
lets do that way, unless @Merisho wants to add some more context fast :-)

JKRhb:
> lets do that way, unless @Merisho wants to add some more context fast :-)

:D I would close for now, let us know, @Merisho, if you want to address this topic again :)

PR #178, opened 2018-05-07
Implemented Block1 transfers. Fixed dgram socket being closed prematurely

Jamezo97:
# Type-1 Block-wise transfer
Implementation of type-1 block-wise transfers. (Based on RFC7959)
Most of the logic has been implemented in 'segmentation.js'
This was implemented for testing purposes for another project, however may be useful for others too.
It has been tested against the Californium CoAP framework without any issue.
There's some examples in 'blockwise_put.js'

# DGRAM Socket closing prematurely
It seems that in rare circumstances, the CoAP agent is flagged to be closed, and whilst in the process of closing the DGRAM socket (between '_cleanUp' and '_doClose') there exists a small window where another CoAP request can be made.
Even if this request is made, the socket is still closed, and the response to the request is never received.
Resetting the '_closing' flag when a new request is made seems to solve the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,6 +312,13 @@ Since v0.7.0, this library supports blockwise transfers, you can trigger
 them by adding a `req.setOption('Block2', new Buffer([0x2]))` to the
 output of [request](#request).
 
+And since v?.?.?, this library supports rudimentry type 1 blockwise transfers, you can trigger
+them by adding a `req.setOption('Block1', new Buffer([0x2]))` to the
+options of [request](#request).
+
+(The hex value 0x2 specifies the size of the blocks to transfer with.
+Use values 0 to 6 for 16 to 1024 byte block sizes respectively.)
+
 See the
 [spec](http://tools.ietf.org/html/draft-ietf-core-coap-18#section-5.4)
 for all the possible options.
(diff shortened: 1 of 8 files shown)


everhardt:
I think it would be wise to also add an expiration timeout like is done with the block2 cache, otherwise clients can fill up the block1 cache by doing a lot of requests for which they do not send in all blocks.

Jamezo97:
> I think it would be wise to also add an expiration timeout like is done with the block2 cache, otherwise clients can fill up the block1 cache by doing a lot of requests for which they do not send in all blocks.

Yep, good idea. I've done so, and separated the caching implementation into a separate file.
I've also cleaned up a few issue with the block1 transfer, specifically if the server servicing the request took too long, it would send the incorrect ACK back and result in a 2.31 response code (continue) instead of whatever was specified by the resultant response object.

Apollon77:
@Jamezo97 This PR needs a rebase on current master and also test are completely missing for you new functionality? Can you please add them? Thank you!

ats-org:
Any possible update on this?

Jamezo97:
Hi all. Sorry for abandoning this.
Looks like you've got things well sorted over at #233 .
I'll close this PR

PR #181, opened 2018-05-09
Update travis tests

neophob:
use node 10 instead node 9


neophob:
make sense, will update pr

neophob:
done

Issue #168, opened 2018-02-28
Malformed response triggers uncatchable error, crashing the calling program


phretor:
If the option length is > 16 bits, the following line throws an exception:

https://github.com/mcollina/node-coap/blob/master/lib/option_converter.js#L140

```
var contentFormatToString = function(value) {
  if (value.length === 0)
    return formatsBinaries[0]

  if (value.length === 1)
    value = value.readUInt8(0)
  else if (value.length === 2)
    value = value.readUInt16BE(0)
  else
    throw new Error('Content-Format option is too big')

  var result = formatsBinaries[value]

  if (!result)
    throw new Error('No matching format found')

  return result
}
```

Because of how `contentFormatToString` is called, I can't find a way to catch that error in the calling program.

If I understand the RFC correctly, the max length is indeed 2 bytes, so the error is worth throwing. However, there must be a way to properly allow the calling program to catch it (e.g., and decide to skip the packet).

I'm proposing the following ugly patch in `incoming_message.js`:

```
function IncomingMessage(packet, rsinfo, outSocket) {
  Readable.call(this)

  try {
    pktToMsg(this, packet)
  } catch (e) {
    //any idea on how to emit a proper event here?
  }

  this.rsinfo = rsinfo
  this.outSocket = outSocket

  this._packet = packet
  this._payloadIndex = 0
}
```

Am I overlooking something?

mkovatsc:
Applications are also forced to crash when a server sends an unknown/unregistered Content-Format number, as the error rooted in `UDP.onMessage` cannot be caught.

Also need proper error event for this, so it does not force-crash the application.

GiedriusM:
IMO any incoming packet handling exceptions should be caught and ignored. What do you guys think?

mkovatsc:
I was thinking about an error handler that could be defined on Agent to see errors and if it is not defined, then maybe just ignore them. However, then there will be open requests that never receive a response, e.g., because the application did not know about the Content-Format beforehand (which would be totally fine for generic clients). So there should also be an error handler on the request message to deal with that.

neophob:
error handler sounds good to me

mkovatsc:
I currently only have the client side of the issue. More thoughts are needed for server-side incoming errors that originate in UDP.onMessage.

For all socket-level errors for clients, the Agent is probably the best place to catch. 

For message-level errors (i.e., no useful response IncomingMessage can be created), the OutgoingMessage (i.e., request) should be notified on its `error` handler.

[...]

PR #194, opened 2018-09-17
Propagating malformed packet errors up to the caller through 'error' …

phretor:
This should partially address #168


phretor:
I'll try the approach suggested by @mkovatsc

PR #196, opened 2018-09-25
We let client handle invalid formats as per #168

phretor:
Based on the discussion on #168 - although it's worth throwing an error when content format is invalid, because how the library is invoked (callbacks), thrown errors can't be caught by the caller, causing the main program to just crash.

Here, we're relaxing a bit the enforcing of the RFC, and let the calling client decide what to do when the response data does not match any valid/known format. For this latter case, we're passing the raw value instead of throwing.


Apollon77:
Please also adjust the tests, currently this PR is not mergable

Apollon77:
@phretor Please adjust the tests accordingly to your changes. Thank you!

test/server.js
```diff
@@ -337,7 +337,7 @@ describe('server', function() {
         var response = parse(msg);
 
         expect(response.code).to.equal('5.00')
-        expect(response.payload.toString()).to.equal('No matching format found')
+        // expect(response.payload.toString()).to.equal('No matching format found')
```
JKRhb, line 340: This line prevented the tests from succeeding. However, the transmitted payload (`assert is not defined`) is very weird and indicates that there is something wrong with the new solution.

lib/option_converter.js
```diff
@@ -137,12 +137,13 @@ var contentFormatToString = function(value) {
...
   else
-    throw new Error('Content-Format option is too big')
+    return value    // the application is still responsible to validate the value
+    //throw new Error('Content-Format option is too big')
 
   var result = formatsBinaries[value]
 
-  if (!result)
-    throw new Error('No matching format found')
+  if (!result) return value
```
JKRhb, line 145: I am wondering if it makes sense to return the value here or if it would be better to return `null` since a value that is longer than two bytes is prohibited by the RFC. Using `null` would probably make the handling of errors easier as well I assume?

JKRhb (commented):
I took the freedom of merging the master branch into this PR branch, finding the problem with the failing test. By doing so I came across two issues (outlined below) that I think should be addressed before merging this PR.

JKRhb:
After investigating the problem a bit more, I now think it would be better to add a try-catch-block to `agent.js` as this is the place where the error causes the crash or the `packetToMessage` function. The change originally proposed in this PR has too many unintended side-effects imho.

test/server.js
JKRhb: I now realized that the new error message is coming from the `assert` in line 346 that is (obviously) undefined. I am wondering if requests like this should rather be handled normally or if a `4.15` ("unsupported content-format") should be returned.

lib/option_converter.js
JKRhb: After thinking about this, returning the numeric value in this very case actually makes a lot of sense to ensure that all messages can be handled. However, maybe returning `null` when the content-format is too big could be an option?

lib/option_converter.js
```diff
@@ -137,12 +137,12 @@ var contentFormatToString = function(value) {
   else if (value.length === 2)
     value = value.readUInt16BE(0)
   else
-    throw new Error('Content-Format option is too big')
+    return null
```
JKRhb, line 140: In the case of an illegal format format, `null` is now returned which is handled well by the rest of the implementation. However, using the raw value or returning a `4.15` are also still on the table as options.

Apollon77: 4.15 would be a clean error ... feels better then null ... but honestly you are deeper in the specs/code then me right now

JKRhb: Yeah, you are right. I think then it might be best to keep the code in this file as is, add another try-catch-block to `agent.js` (which is the part of the library that is crashing at the moment) and treat the content-format as `text/plain` if an illegal value is encountered. The server on the other hand could return `4.15` instead of `5.00` when it catches the exception. 

I think another part of the solution would be to increase the number of content-formats that are handled by default... I will have another look into this over the next couple of days :)

JKRhb: I had another look into the [RFC](https://datatracker.ietf.org/doc/html/rfc7252#section-3) and noticed that the content-format is an "elective" option that has to be "silently ignored". I therefore changed the behavior of the agent and the server to either use the numeric value or to simply drop the option if its format is an "illegal" one. This is now also reflected in the test cases.

Regarding the additional options I would open another PR soon(ish) so that we could maybe also include this PR in the upcoming release and make the library a bit more stable :)

Apollon77: makes sense

JKRhb:
@Apollon77 If it's okay for you then I would merge this PR into master

JKRhb:
I created #248 as a follow-up for more content-formats.

Apollon77:
> @Apollon77 If it's okay for you then I would merge this PR into master

Go ahead, ok for me!

Issue #202, opened 2019-02-25
RetrySend does not work properly


blankm:
Somehow retrySend is buggy.
When I have an server being offline and send a request to it, the timeout in RetrySend.prototype.send properly fires. 
When I then start my server and rerun the script, all's fine.
Strangely, if I kill the server after the first successful request (or later), the timeout does not fire anymore, the agent simply stops and does nothing. 
Even more confusing, when I debug the scenario with the server being killed after the first successful request (or later), the timeout works again?

But due to the rare comments and the way calls are made when sending (and resending) messages and the different timers, I wasn't able to fully understand the code and identify the part that causes this strange behavior.

GiedriusM:
Hi @blankm, 
Could you clarify how many requests do you send and when the server is stopped? If you could provide some code snippets that recreate the problem that would help too.

blankm:
Basically it's running in an endless loop. The request is always awaited, so the blocking one stops the whole thing. So, what I do to trigger the problem:
I have a coap server running which offers a value. I have a script (see below) that reads that value in a interval (the interval period doesn't have any influence). Once I kill the coap server, the script stops forever at the `await`, no timeout or error is fired. Strangely, if I step through the code, but apart from that, everything the same like written above, an error is emitted. And, as said, if I start the script without the coap server running, also an error is fired immediately. 

It looks like this:

- Init:

```
    this.agent = new coap.Agent();
    
    let coapTiming = {
      ackTimeout:0.25,
      ackRandomFactor: 1.0,
      maxRetransmit: 3,
      maxLatency: 2,
      piggybackReplyMs: 10
    };
    coap.updateTiming(coapTiming);
```

- Inside my loop (only the coap stuff), I call this function readResource and wait to finish with `await`:

``` 
function myLoopWhichIsCalledByATimer(){
   return await readResource("coap://127.0.0.1:5683/dummy");
}
```

```
public readResource(uri): Promise<Content> {
    return new Promise<Content>((resolve, reject) => {

      let requestUri = url.parse(uri);
      let options: CoapRequestConfig = {
        agent: this.agent,
        hostname: requestUri.hostname,
        port: parseInt(requestUri.port, 10),
        pathname: requestUri.pathname,
        query: requestUri.query,
        observe: false,
        multicast: false,
        confirmable: true
      };
      options.method = 'GET';
   
      let req = this.agent.request(options);
      

      req.on("response", (res: any) => {
[...]

PR #204, opened 2019-03-01
Fix: Retrysend broken when passing socket to agent

blankm:
Fixes issue #202

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -58,6 +58,9 @@ function Agent(opts) {
 util.inherits(Agent, events.EventEmitter)
 
 Agent.prototype._init = function initSock(socket) {
+  
+  this._closing = false;
+  
   if (this._sock) {
     return
   }
@@ -99,8 +102,7 @@ Agent.prototype._init = function initSock(socket) {
 
   this._lastToken = Math.floor(Math.random() * (maxToken - 1))
   this._lastMessageId = Math.floor(Math.random() * (maxMessageId - 1))
-
-  this._closing = false
+  
   this._msgInFlight = 0
   this._requests = 0
 }


GiedriusM:
Thank you for finding the issue :) Could you check if adding `this._closing = false` to the `_doClose()` function solves the issue? To me it looks like it's the "deinitialization" problem, which leaves an invalid state, while the code in the init is supposed to run only once. IMO the fix should reflect that.

Also did you find a way to consistently reproduce the bug? It would be nice to have a test to catch this in the future :)

blankm:
> Thank you for finding the issue :) Could you check if adding `this._closing = false` to the `_doClose()` function solves the issue? To me it looks like it's the "deinitialization" problem, which leaves an invalid state, while the code in the init is supposed to run only once. IMO the fix should reflect that.

The init is with the code provided meant to be run every time a new request is created, at least it is called each time a new request is created, so fixing the bug there (and there it actually is located) makes the most sense for me. The cleanup in this stack was written the way it is, so it's definitely not a bug of the deinitialization, since this behaves as intended by the author. The init does not behave as intended (when using externally provided sockets). One could discuss whether the way, this stack handles init and deinit could be done in another way, but that's not the point here. This "single-line-bugfix" solves the issue.

> 
> Also did you find a way to consistently reproduce the bug? It would be nice to have a test to catch this in the future :)
Yeah, provide an external socket , create a request to a server that exists, let it finish and then create a request to a sever that doesn't exist, the second request's promise will leave you wait forever. But I don't know whether this can be automated.

blankm:
Are there plans to merge this PR? I am eagerly waiting for the npm package being updated with this bugfix, so that I can update my project to use the new version of the node-coap package.

GiedriusM:
I would like to have a test case for this issue. So far all my attempts to reproduce it worked as expected.
For example:
```javascript
const coap = require('coap');
const dgram = require('dgram');

let coapTiming = {
  ackTimeout:0.25,
  ackRandomFactor: 1.0,
  maxRetransmit: 3,
  maxLatency: 2,
  piggybackReplyMs: 10
};
coap.updateTiming(coapTiming);

const server = coap.createServer();
server.on('request', (req, res) => {
  res.end();
});
server.listen(7892);

const sock = dgram.createSocket('udp4');
sock.bind();

const agent = new coap.Agent({ socket: sock });

const req = coap.request({ agent: agent, port: 7892 });
req.end();
req.on('response', (res) => {
  console.log('Req1 response');
  setImmediate(() => {
    const req2 = coap.request({ agent: agent, port: 7893 });
    req2.end();
  });
});
```
does emit timeout.

Looking into the agent, I can see that the `_closing` flag should not be set (to true) when an external socket is used:
```javascript
  if (!this._opts.socket)
    this._closing = true
```
How is `_closing` flag set in your case?

blankm:
Without having a look at the code in detail, your idea seems to be valid, too.
I fixed it in
`Agent.prototype._init = function initSock(socket)`
since, the original code (I guess at the time, when no external sockets where used) set 
`this._closing` to false again there.
Your solution is valid if `this._closing` is only used for the purpose of closing the socket. But I have the code not in mind right now, so I can't say whether setting `this._closing` to like you suggest has any interference.

GiedriusM:
> Yeah, provide an external socket , create a request to a server that exists, let it finish and then create a request to a sever that doesn't exist, the second request's promise will leave you wait forever. But I don't know whether this can be automated.

Could we focus on reproducing this in a test case? The problem is still confusing me.

blankm:
Any updates here?
Testcase can be set up using a CoAP server that is killed after the first sucessful request, so a subsequent request should cause a timeout which it doesnt in the old code.

GiedriusM:
> Testcase can be set up using a CoAP server that is killed after the first sucessful request, so a subsequent request should cause a timeout which it doesnt in the old code.

server.js:
```javascript
const coap = require('coap');

let counter = 0;

const server = coap.createServer();
server.on('request', (req, res) => {
  res.end('' + counter++);
});
server.listen(7892);
```

client.js:
```javascript
const coap = require('coap');
const dgram = require('dgram');

let coapTiming = {
  ackTimeout:0.25,
  ackRandomFactor: 1.0,
  maxRetransmit: 3,
  maxLatency: 2,
  piggybackReplyMs: 10
};
coap.updateTiming(coapTiming);

const sock = dgram.createSocket('udp4');
sock.bind();

const agent = new coap.Agent({ socket: sock });

function doRequest(res) {
  if (res) {
    console.log('Got response', res.payload.toString());
  }

  const req = coap.request({ agent: agent, port: 7892 });
  req.end();
  req.on('response', doRequest);
  req.on('timeout', () => { console.log('Timeout') });
  req.on('error', () => { sock.close() });
}


doRequest();
```

When I run only client.js I get a timeout.
When I run server.js, then client.js and then kill (ctlr+c) server.js I get a timeout.

This and all my previous attempts to reproduce the problem based on your description seems to have failed. Could you provide an example code that causes the issue or even better a complete test case for travis?

kevinresol:
I realize the "bug" will occur when there is no explicit calling of `updateTiming()`. I think there should be some sensible defaults?

```js
// server.js
const coap = require('coap');

let counter = 0;

const server = coap.createServer();
server.on('request', (req, res) => {
	res.end('' + counter++);
});
server.listen(7892);
```

```js
// client.js
var coap = require('coap');

(function loop() {
	console.log('loop');
	var req = coap.request('coap://localhost:7892');
	req.on('response', () => setTimeout(loop, 1000));
	req.on('timeout', () => console.log('Timeout'));
	req.on('error', () => console.log('Error'));
	req.end();
})();
```

Steps: 
1. start server
2. start client
3. kill server

JsonMa:
@Apollon77 Can you reproduce this problem? I can't reproduce it locally..

Apollon77:
No I can‘t but the code change makes sense logically and when reading the issue and the Pr the fix seems to be proofed by users. Additionally it does not really break things so risk is ok. 

I learned that I can not repro anything and so I review things a bit different in these cases and more do a logical due dilligence. ;-)
Yes ideally each issue also adds a test that breaks without the test and then fixing it ... but sometimes this is just not possible.

Apollon77:
@mcollina What do you thingk on how to handle such edgy cases like this?

mcollina:
Unfortunately I do not have time to dig in.

Apollon77:
I meaned more the general "approach" question ... ok to accept it if there was a proof by usersand code "makes sense" also if we can not really repoduce it? I would accept it ion such situations

JsonMa:
@Apollon77 OK，I agree with it. And if someone approve this change would result in another bug, we revert it.

Issue #205, opened 2019-03-26
Multiple responses to a multicast request


mateusz-:
Hi,

The README.md states the following:
```markdown
multicast: (...) Several response events will be emitted for each received response.
```

This does not seem to be the case. The callback gets called only once, from what I can see.

Here's a Wireshark summary:
```
1 0.000 fdde:ad00:beef::b2f8:7526:d5a2:6f51   CoAP    NON, MID:32309, GET, TKN:50 53 1c d5, /v1/info
2 0.046 fdde:ad00:beef::b57c:7833:400:0       CoAP    NON, MID:10036, 2.05 Content, TKN:50 53 1c d5, /v1/info
3 0.048 fdde:ad00:beef::15d2:fce4:0:d000      CoAP    NON, MID:16973, 2.05 Content, TKN:50 53 1c d5, /v1/info
4 0.054 fdde:ad00:beef::b2f8:7526:d5a2:6f51   CoAP    RST, MID:16973, Empty Message
```

Frame 1 is a multicast request to `ff03::2`. Frame 2 and 3 are responses. Frame 4 is generated as a response to frame 3 (i.e., it is sent to `fdde:ad00:beef::15d2:fce4:0:d000`) by node-coap. This should not be happening (RST), unless a timeout was reached (it hasn't). The `response` callback gets called once for response from `fdde:ad00:beef::b57c:7833:400:0` (frame 2).

My JavaScript looks like this:
```js
const controllers = {};

const req = coap.request({
  host: 'ff03::2',
  pathname: '/v1/info',
  port: 3002,
  method: 'GET',
  confirmable: false,
  retrySend: 0,
  multicast: true,
  multicastTimeout: 2000,
});

req.on('response', res => {
  try {
    const controller = JSON.parse(res.payload.toString());
    controllers[controller.id] = controller;
  } catch(err) {
    debug.error('CoAP GET /v1/info error', err);
  }
});

req.end();

setTimeout(() => {

  res.data.msg = controllers;

  next();

}, 2001);
```

I created a [fix to this in my own fork](https://github.com/mateusz-/node-coap/commit/53f191e93202da72eabdbe6be646ee2c19070bc9), haven't created a pull request yet though. All tests pass. I did think about creating a new test for this, but this would require multiple servers listening on the same port that handle a multicast request. These servers would have to be mocked up, I believe. Currently, servers are not mocked.

If my fix looks good, I'll create a pull request.

GiedriusM:
[...]

PR #206, opened 2019-03-27
Fixes multiple responses to a non-confirmable multicast request

mateusz-:
See issue #205.

The idea here is to remove the token-to-request mapped value after the multicast timer times out. Currently, the token is removed once the first response is handled, which prevents other responses from being handled.

Added a test with two servers handling a request. Each server responds with a different MID, thus the response is not mapped using the mid-to-requst map.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -252,7 +252,7 @@ Agent.prototype._handle = function handle(msg, rsinfo, outSocket) {
   else if (block2) {
     delete that._tkToReq[req._packet.token.readUInt32BE(0)]
   }
-  else if (!req.url.observe && packet.token.length > 0) {
+  else if (!req.url.observe && packet.token.length > 0 && !req.multicast) {
     // it is not, so delete the token
     delete that._tkToReq[packet.token.readUInt32BE(0)]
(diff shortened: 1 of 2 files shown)


GiedriusM:
I would prefer second eyes to check the PR. @neophob could you do it? If not I'll go with the merge.

lib/agent.js
```diff
@@ -377,6 +377,7 @@ Agent.prototype.request = function request(url) {
   // Start multicast monitoring timer in case of multicast request
   if (url.multicast === true) {
     req.multicastTimer = setTimeout(function() {
+      delete that._tkToReq[req._packet.token.readUInt32BE(0)]
```
mateusz-: Noticed that `req._packet.token` is undefined if `end()` is never called on the response. Once the timeout function is invoked, `readUInt32BE()` is called on the `undefined` token and an exception is thrown.

GiedriusM:
I would like to merge this PR. @mateusz- are we good to go, i. e. no other pending changes?

mateusz-:
Yes, it's good to go.

PR #211, opened 2019-10-17
Fix more flag during block-wise transfer

JcBernack:
If the payload was an exact multiple of the block size the index calculation was off by one.
This caused clients to make an additional redundant request at the end.
Also, GET requests would always have the more flag set, which was against the specs.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -183,7 +183,7 @@ create buffer for block2 option
 module.exports.createBlock2 = function createBlock2(requestedBlock) {
   var byte
   var szx = Math.log(requestedBlock.size)/Math.log(2) - 4
-  var m = ((requestedBlock.moreBlock2==true)?0:1)
+  var m = ((requestedBlock.moreBlock2==true)?1:0)
   var num = requestedBlock.num
   var extraNum
 
--- a/test/blockwise.js
+++ b/test/blockwise.js
@@ -165,7 +165,6 @@ describe('blockwise2', function() {
         port: port
     })
     .setOption('Block2', Buffer.of(0x3D)) // request for block index 3
-
     .on('response', function(res) {
       expect(res.code).to.eql('4.02')
       //expect(cache.get(res._packet.token.toString())).to.be.undefined
--- a/test/blockwise.js
+++ b/test/blockwise.js
@@ -19,7 +19,7 @@ describe('blockwise2', function() {
     , clientPort
     , client
     , bufferVal
-    , payload   = new Buffer(1536)
+    , payload   = Buffer.alloc(1536)
 
   beforeEach(function(done) {
     bufferVal = 0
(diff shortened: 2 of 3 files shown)


test/blockwise.js
```diff
@@ -19,7 +19,7 @@ describe('blockwise2', function() {
     , clientPort
     , client
     , bufferVal
-    , payload   = Buffer.alloc(1300)
+    , payload   = new Buffer(1536)
```
JcBernack, line 22: Update to new Buffer API, see #212

test/blockwise.js
```diff
@@ -158,10 +158,14 @@ describe('blockwise2', function() {
...
 
   it('should receive error request for out of range block number', function(done) {
+    // with a block size of 512 and a total payload of 1536 there will be 3 blocks
+    // blocks are requested with a zero based index, i.e. indices 0, 1 and 2
+    // block index 3 or higher is "out of range" and should cause an error response
     var req = coap.request({
         port: port
     })
-    .setOption('Block2', Buffer.of(0x55)) // request for block 5, size = 512 from 1300B msg (total 1300/512=3 blocks)
+    .setOption('Block2', new Buffer([0x3D])) // request for block index 3
```
JcBernack, line 167: See above.

lib/helpers.js
```diff
@@ -183,7 +183,7 @@ create buffer for block2 option
 module.exports.createBlock2 = function createBlock2(requestedBlock) {
   var byte
   var szx = Math.log(requestedBlock.size)/Math.log(2) - 4
-  var m = ((requestedBlock.moreBlock2==true)?0:1)
+  var m = ((requestedBlock.moreBlock2==true)?1:0)
```
JsonMa, line 186: `requestedBlock.moreBlock2` is a boolean, maybe this statement can be abbreviated as `var m = requestedBlock.moreBlock2 ? 1:0`.

JsonMa:
@JcBernack Thanks for your PR and could you please rebase the commit and squash the whitespace commit?

lib/helpers.js
Apollon77: I think we should simply do this change if relevant here or leave this PR out of the current fix version.
In the end I'm not sure that you can assume it is a boolean, but in fact the change you suggest also fix other "truthy" cases and is equivalent

JcBernack: Leaving out the `== true` is exactly equivalent, imho. If `moreBlock2` is not boolean JS does its coercion magic as this is not a `===`. I just didn't change it because I wanted to touch the least amount of code necessary.

JsonMa: @JcBernack I do understand your point, in fact requestedBlock in this sentence will be processed as `!!requestedBlock.moreBlock2`, so event though the value is not boolean, like `undefined` it would be still be converted to a boolean. But in this case, we can pr it first and submit another pr to optimize it.

Apollon77:
@JsonMa Ready to merge for me, it is right to fix bug without introducing "logic behaviour" changes ;-)

Issue #195, opened 2018-09-24
Observe cancellation by client


mkovatsc:
node-coap currently does not support proactive cancellation via GET with Observe=1.

In `server.js`, the `_handle()` function should probably try to find the corresponding ObserveStream via the token when `request.headers['Observe'] === 1` instead of producing a new response with ` response = new Message(packet, function(response, packet)`. The ObserveStream needs to be closed. Sending the last (cached) response/notification with that token should work; otherwise, ObserveStream could also take a callback to produce a final response.

lucarv:
I wonder how to implement this correctly. According to (my perhaps poor) understanding of the specs, when the client stop sending **ack** to a **data** posted event, the server should then close the stream. I am not sure how to stop the client from sending an **ack** to the **data** event. Is it even possible?

sjlongland:
Well, no, not when it stops sending `ACK`s… that'll just cause the server to re-send the notification until it exhausts the retry count.

There are two ways:
* Client replies to the next notification with `RST`.
* Client sends a `GET` with token matching the original request and `Observe=1`.

`node-coap` does the former when you close the stream, it just cleans up the request at the client end, so next notification it naturally responds with the `RST` message.  To be proactive about things, it just needs to generate that `GET` request at that time instead.

GiedriusM:
Per RFC7641 the default way of handling a canceled observation is by sending `RST`:

> A client that is no longer interested in receiving notifications for a resource can simply "forget" the observation.`

and the active cancellation is optional:

> a client MAY explicitly deregister by issuing a GET request that has the Token field set to the token of the observation to be cancelled and includes an Observe Option with the value set to 1 (deregister). 

A flag to the close method could be added to indicate active/forced cancellation:

```javascript
ObserveReadStream.prototype.close = function(sendCancel) {
  if (sendCancel) {
    // duplicate original req and send with observe=1 ...
  }
  this.push(null)
  this.emit('close')
}
```

However,

> ... All other options MUST be identical to those in the registration request except for the set of ETag Options.

and the whole request option duplication may look a bit messy/out-of-context in this place.

Your thoughts?

lucarv:
@sjlongland my bad, i should have re-read the RFC instead of just googling it ;) I found something [here](http://programmingwithreason.com/article-iot-coap.html) and believed it blindly. Anyway, I suppose the lib does not support Observe set to false, and the reset() crashes (there is an issue on it). I found a way around anyway, thanks for the reply.

[...]

PR #214, opened 2020-03-17
Add support for de-registering observations

sjlongland:
This is a superset of pull request #164, bringing forward and building on the work of @pekkanikander.  Since this alone didn't go all the way to solving issue #195, I also bit the bullet and implemented that too.

Due to the commit hooks set up on this repository, I was forced to fix a few other issues as part of this pull request, notably the "leaking" of exceptions by some callback functions, as well as extending the time-outs on some unhappy path tests that were failing simply because some network operations were taking a bit longer to signal an error.

Finally, since it's required for RFC-7641, I have added the ability for a user to specify their own token, that way when de-registering, we can honour the spec and send the correct token to the CoAP server.

Since the de-registration would _not_ have a token matching `_lastToken` anymore, it was necessary to tweak how requests were matched against their tokens: we now convert the token to `hex` representation, which has the bonus of removing the 4-byte hard limit on token lengths present before.

I have tested this implementation against the [OpenThread CoAP server](https://github.com/openthread/openthread/pull/4396), and it seems to perform as expected.

### DIFF
--- a/lib/observe_read_stream.js
+++ b/lib/observe_read_stream.js
@@ -47,9 +47,12 @@ ObserveReadStream.prototype.append = function(packet) {
   }
 }
 
-ObserveReadStream.prototype.close = function() {
+ObserveReadStream.prototype.close = function(eagerDeregister) {
   this.push(null)
   this.emit('close')
+  if (eagerDeregister) {
+    this.emit('deregister');
+  }
 }
 
 // nothing to do, data will be pushed from the server
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -380,10 +380,14 @@ Agent.prototype.request = function request(url) {
     var token
     if (!(packet.ack || packet.reset)) {
       packet.messageId = that._nextMessageId()
-      if (url.token instanceof Buffer)
+      if ((url.token instanceof Buffer) && (url.token.length > 0)) {
+        if (url.token.length > 8) {
+          return req.emit('error', new Error('Token may be no longer than 8 bytes.'));
+        }
         packet.token = url.token;
-      else
+      } else {
         packet.token = that._nextToken()
+      }
       packet.token = that._nextToken()
       token = packet.token.toString('hex')
       that._tkToMulticastResAddr[token] = []
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -9,6 +9,7 @@
  */
 
 var util                  = require('util')
+  , crypto                = require('crypto')
   , events                = require('events')
   , dgram                 = require('dgram')
   , parse                 = require('coap-packet').parse
@@ -333,12 +334,13 @@ Agent.prototype._handle = function handle(msg, rsinfo, outSocket) {
 }
 
 Agent.prototype._nextToken = function nextToken() {
-  var buf = Buffer.alloc(4)
+  var buf = Buffer.alloc(8)
 
   if (++this._lastToken === maxToken)
     this._lastToken = 0
 
   buf.writeUInt32BE(this._lastToken, 0)
+  crypto.randomBytes(4).copy(buf, 4)
 
   return buf;
 }
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -315,6 +315,16 @@ Agent.prototype._handle = function handle(msg, rsinfo, outSocket) {
       }
       that._cleanUp()
     })
+    response.on('deregister', function() {
+      var deregister_url = Object.assign({}, req.url);
+      deregister_url.observe = 1;
+      deregister_url.token = req._packet.token;
+
+      let deregister_req = that.request(deregister_url);
+      // If the request fails, we'll deal with it with a RST message anyway.
+      deregister_req.on('error', function() {});
+      deregister_req.end();
+    })
   } else {
     response = new IncomingMessage(packet, rsinfo, outSocket)
   }
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -380,6 +380,10 @@ Agent.prototype.request = function request(url) {
     var token
     if (!(packet.ack || packet.reset)) {
       packet.messageId = that._nextMessageId()
+      if (url.token instanceof Buffer)
+        packet.token = url.token;
+      else
+        packet.token = that._nextToken()
       packet.token = that._nextToken()
       token = packet.token.toString('hex')
       that._tkToMulticastResAddr[token] = []
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -148,8 +148,8 @@ Agent.prototype._handle = function handle(msg, rsinfo, outSocket) {
         }
       }
   if (!req) {
-    if (packet.token && packet.token.length === 4) {
-      req = this._tkToReq[packet.token.readUInt32BE(0)]
+    if (packet.token.length > 0) {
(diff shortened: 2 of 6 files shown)


sjlongland:
Not quite sure why that last test run just failed in NodeJS 8 and nothing else.  Looks like a timing glitch, *and* another uncaught exception.

sjlongland:
I'd suggest an audit done of the test cases, any "inner" call-back function that calls `expect`, there should be a `try`-`catch` block there that calls `done(err)` so the test bails properly instead of giving false positives like the ones seen in this PR.

I think I've caught the ones that have bitten me here, but there are plenty of others.  I think though it's getting beyond the scope of this PR to fix them.

Apollon77:
@sjlongland Could you please rebase your PR on master?! Thank you!

sjlongland:
@Apollon77 No problems… I just tried a quick rebase now, haven't tested this version out against other CoAP implementations to see if it "works" yet, if I get some time I'll try to give it a test and see how it fares.

sjlongland:
```
3	0.078498	fd34:fe56:7891::1	fd34:fe56:7891:12:58bd:dd1f:ac3a:912d	CoAP	72	CON, MID:17172, GET, TKN:45 42 36 5f, /sgw?t=3D9UCD44
…
24	7.899396	fd34:fe56:7891::1	fd34:fe56:7891:12:58bd:dd1f:ac3a:912d	CoAP	73	CON, MID:17175, GET, TKN:45 42 36 5f, /sgw?t=3D9UCD44
25	7.955430	fd34:fe56:7891:12:58bd:dd1f:ac3a:912d	fd34:fe56:7891::1	CoAP	60	ACK, MID:17175, 2.02 Deleted, TKN:45 42 36 5f, /sgw
```

Just did a test with one of our devices… `fd34:fe56:7891::1` is the server running the `node-coap` client… sending a de-register to the server (OpenThread).

JKRhb:
@Apollon77 @sjlongland Do you think this PR could be merged after another rebase? :)

Apollon77:
I'm a bit unsure about 

>  we now convert the token to hex representation, which has the bonus of removing the 4-byte hard limit on token lengths present before.

... I did not checked the specs if this is a definition to have 4 bytes or if it is realy a limitation ... In the past I remember that we needed to add these length checks in some places, but I do not remember if it was because token as "length ===0) or any other length <4

And measn of "eager" variable name is unclear to me ... https://github.com/mcollina/node-coap/pull/214/files#diff-89afb2a479e2d2d3990267205702b41b6fe6af1f7cee42223091692ccd67e934R50 ...I can not bind any meaning to that that would bring me to "we deregisterr on close" :-)

sjlongland:
> I'm a bit unsure about

> > we now convert the token to hex representation, which has the bonus of removing the 4-byte hard limit on token lengths present before.

> ... I did not checked the specs if this is a definition to have 4 bytes or if it is realy a limitation ... In the past I remember that we needed to add these length checks in some places, but I do not remember if it was because token as "length ===0) or any other length <4

So, the original code used `readUInt32BE` to parse the first 4 bytes as a 32-bit unsigned integer in big-endian mode, then used the value as the key in `_tkToReq`.  The thing is, the `token` field in a CoAP message is meant to be an opaque value, not meaningful to anybody, [can be up to 8 bytes in length](https://datatracker.ietf.org/doc/html/rfc7252#section-5.3.1), and each byte may be any value.

> And measn of "eager" variable name is unclear to me ... https://github.com/mcollina/node-coap/pull/214/files#diff-89afb2a479e2d2d3990267205702b41b6fe6af1f7cee42223091692ccd67e934R50 ...I can not bind any meaning to that that would bring me to "we deregisterr on close" :-)

The language there comes directly from RFC-7641:
>    In some circumstances, it may be desirable to cancel an observation
>    and release the resources allocated by the server to it more **eagerly**.
>    In this case, a client MAY explicitly deregister by issuing a GET
>    request that has the Token field set to the token of the observation
>    to be cancelled and includes an Observe Option with the value set to
>    1 (deregister).  All other options MUST be identical to those in the
>    registration request except for the set of ETag Options.  When the
>    server receives such a request, it will remove any matching entry
>    from the list of observers and process the GET request as usual.

So we give the user a choice, we can simply forget about it on our end, and wait for the server to send the next observation and send `RST` in reply, *lazy* de-registration, or we can be explicit about it, and de-register it when we know we no longer want it, *eager* de-registration.

lib/observe_read_stream.js
```diff
@@ -47,9 +47,12 @@ ObserveReadStream.prototype.append = function(packet) {
   }
 }
 
-ObserveReadStream.prototype.close = function() {
+ObserveReadStream.prototype.close = function(eager) {
```
JKRhb, line 50:
```suggestion
ObserveReadStream.prototype.close = function(eagerDeregistration) {
```

lib/observe_read_stream.js
```diff
@@ -47,9 +47,12 @@ ObserveReadStream.prototype.append = function(packet) {
...
   this.push(null)
   this.emit('close')
+  if (eager) {
```
JKRhb, line 53:
```suggestion
  if (eagerDeregistration) {
```

JKRhb (commented):
Maybe `eager` could simply be renamed to make the purpose of the parameter more clear? Otherwise very useful changes as far as I can tell.

Apollon77:
@sjlongland Many thnks for the protocl background infos on both! What do you think of the name proposal from @JKRhb ?

Apollon77:
SP: And the to-hex thing I would then also simply do - so the library was limiting kind of and the change cleans it up ... Formally a check foor the lentth could help, but honestly ... more flexible that way I think. maybe we could add a testcase with a 8 byte number just to show that it also wirorks with >4 and we did not forgot anything where code still checks for a 4

sjlongland:
@Apollon77 Yes, I've applied @JKRhb's variable name suggestion, slipped that into the rebase at 0d34a94 (although I note I should have changed the commit subject… never mind).  I also realised there was a loophole where someone could specify `Buffer.from([])` as their "token"… so added a check for `length=0` and ignore the user-supplied token in that case.

I'll see what I can do about adding a couple of test cases… both the `length=8` case, and a `length=0` case to make sure I closed-off that loophole.

Apollon77:
@sjlongland Thank you, that would be awesome

sjlongland:
So, there was already a `length=8` test case, seemingly working just fine.

The very last commit actually changes the _default_ token length to 8 bytes by adding random bytes to the end of the 4-byte token, so first 4 bytes are sequential as before, but the last 4 are pseudorandom.  That pretty much tests the whole suite against this new behaviour.

JKRhb:
> 
> The very last commit actually changes the _default_ token length to 8 bytes by adding random bytes to the end of the 4-byte token, so first 4 bytes are sequential as before, but the last 4 are pseudorandom. That pretty much tests the whole suite against this new behaviour.

Great, I think that is a very good addition in terms of security! As all of the tests are passing and coverage stayed roughly the same I think this should be ready to be merged :)

@Apollon77 Do you want to hit the button? Or do you see more points that should be addressed?

Apollon77:
Thank you @sjlongland

PR #215, opened 2020-04-08
Update README.md

rmhowie:
Corrected link to IncomingMessage to #incoming instead of #outgoing


rmhowie:
Sorry the description is wrong, it should be:

Corrected link to OutgoingMessage to #outgoing instead of #incoming

PR #219, opened 2020-06-15
fix(agent): fix agent config bug

JsonMa:
Fix the wrapper bug of `new Agent` and remove useless variable in `agent.test.js`


JsonMa:
@mcollina  It would be my pleasure to maintain this project if no one is maintaining. I uses this project very often, cause our company' main product is the Internet of things platform.

PR #220, opened 2020-06-16
handle errors coming from underlying Datagram socket send call

akashgoswami:
Currently, exceptions coming from retry_send (e.g. socket closed) are not being propagated to upper layers.

### DIFF
--- a/lib/retry_send.js
+++ b/lib/retry_send.js
@@ -42,14 +42,18 @@ util.inherits(RetrySend, EventEmitter)
 RetrySend.prototype._send = function(avoidBackoff) {
   var that = this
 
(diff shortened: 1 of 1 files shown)


Apollon77:
Will that really throw an exception synchronous? Do you have the relevant stack trace?

JsonMa:
> Currently, exceptions coming from retry_send (e.g. socket closed) are not being propagated to upper layers.

It doesn't make sense to do that. Also, unit test is missing.

Apollon77:
Yeay and it is also uneeded because the socket in retry class is coming from server.js ... and there an error handler is attached ... I would close that one

PR #222, opened 2020-07-02
change travis testing to test current LTS versions only

Apollon77:


Apollon77:
@JsonMa maybe we should merge this first :-) To make sure we do not get broken builds

.travis.yml
```diff
@@ -3,7 +3,6 @@ sudo: false
 dist: trusty
 before_script: sudo sh -c 'echo 0 > /proc/sys/net/ipv6/conf/all/disable_ipv6'
 node_js:
-  - "6"
-  - "8"
```
JsonMa, line 7: Maybe we should keep this even-numbered release.

Apollon77: No they will break as soon as we update the deps because mocha and chai kicked support for older ones ... I would document in changelog that nodejs 6+8 are not tested any longer and so are not officially suported. They might work but no guarantees

JsonMa: @Apollon77 We can close this PR, cause PR #225 already update these.

PR #225, opened 2020-07-02
update all dependencies, fixes #223

### LINKED ISSUE #223: Update all dependencies
* "devDependencies": {
    * "pre-commit": "1.1.3", ---> why hard pinned? newest is 1.2.2 ... no changelog, no releases, lets hope they do sematic, test it
    * "chai": "^3.4.1", ---> newest 4.2.0, should work
    * "mocha": "^3.2.0", ---> newest 8.0.1, should work
    * "timekeeper": "^1.0.0", --> newest 2.2.0 ... only "freeze" behaviour changed, so should have no effect
    * "sinon": "~1.7.3" ---> newest 9.0.2 ... should work, basically nodejs version changes, test it

*  "dependencies": {
    * "bl": "^1.0.0", ---> newest 4.0.2 ... 2.x added a feature, 3.x removed node 4+5, 4.x added a feature ... should work
    * "capitalize": "^1.0.0", ---> newest 2.0.3 ... is lowercasing rest of string now, used for http headers so should work (https://github.com/mcollina/node-coap/blob/master/lib/helpers.js#L41)
    * "coap-packet": "^0.1.12", ---> 0.1.14, should be easy
    * "debug": "^2.2.0", ---> newest 4.1.1 ... 3.x added features and updated libs, 4.x dropped nodejs 4+5 an change deps. should work too
    * "fastseries": "^1.7.0", ---> newest 2.2.0 ... removed a callback we do not use (releaesed), should work
    * "lru-cache": "^4.0.0", ---> newest 5.1.1 ... haha "remove older node versions" ... no info which ggrrr ... but es6 classes should work node10+, should work
    * "readable-stream": "^2.2.2" ---> newest 3.6.0 ... this is hard to see ... hope tests see it ... need to test @JsonMa  what do you think?

Apollon77:
* Most changes were needed because sinon also mocks setImmediate but it needed to be used oficially and it seems that "Not mocking setImmediate " did never work that good. Maybe at all not the nicest solution, but working


Apollon77:
I personally add the lock file in the meantime becasue then we can enable dependabot and the github security advisories will work, so we can make sure to stay current with deps. it is just here and not in the npm package. I have good experience with it.
Additionally we could use "npm ci" to install for tests which should speedup tests also

mcollina:
Please no package-lock. cc https://github.com/sindresorhus/ama/issues/479

You can have dependabot without a lockfile.

.npmignore
```diff
@@ -1 +1,2 @@
 *~
+package-lock.json
```
mcollina, line 2: no.

Apollon77: Done

JsonMa: We need a .npmrc file to disable package lock.

Apollon77: done

mcollina:
I would recommend to change linting and add a linter in another PR.

Apollon77:
So we need to decide if we want to add it to the "next release" with the other fix PRs already or do two with some days in between.
@JsonMa @mcollina  what do you think?

mcollina:
maybe after that

JsonMa:
@mcollina @Apollon77 I think adding lint  to this project would be a big change.

Apollon77:
@JsonMa might be ... but this would be a different issue then this one ;-))

Apollon77:
> maybe after that

ok, so plan stays

* JSonMa reviews all the relevant fix PRs that are assigned to him except this one :-) I checked them and most I approved already. I think @JsonMa can also directly merge when ok
* Then we (i can do it) create the version bumb PR for the fix version and I would also add a CHANGELOG.md with this
* Then we can relrease this.
* Then we can merge this PR and start also looking into the feature PRs

Apollon77:
oohhh ... ok ... should we revert or go  with a "fixes and deps" release now? ;-)

JsonMa:
> oohhh ... ok ... should we revert or go with a "fixes and deps" release now? ;-)

Sorry, i didn't see your comment... I think no need to revert.

PR #226, opened 2020-07-03
fix(agent): remove useless variables

JsonMa:
These variables have never been used.


JsonMa:
> formally there are also unused "bl" and sinon in test/agent.js ... maybe also remove it there? :-)

They have already been removed in PR #219

Apollon77:
awesome :-)

PR #227, opened 2020-07-03
chore: add nyc and mocha config

JsonMa:
@Apollon77 @mcollina We should add `check-coverage` of nyc to travis, but i don't know the appropriate thresholds, should it be 100%?


Apollon77:
100% is unrealistic ;-)

Yes let’s measure where we are now and try to get higher and not lower ... but it is always a consideration. 

It is more important which parts of the code are covered then which percentage in my eyes.

JsonMa:
image
@Apollon77 @mcollina Current coverage.

Apollon77:
Awesome!! Great work @mcollina ...

JsonMa:
@Apollon77 Can we merge this PR? We need your review...

Apollon77:
Done. Can you please look at the other PRs that we gets fix version out first? And now with this merge (shit misse that) we can not really revert the dep PR because now it is mixed. And now i remember why unsafe that one as „later merge“. 

Ok let’s take the Risk and leave the dep PR in. Else it is getting too complicated. :-(

JsonMa:
> Done. Can you please look at the other PRs that we gets fix version out first? And now with this merge (shit misse that) we can not really revert the dep PR because now it is mixed. And now i remember why unsafe that one as „later merge“.
> 
> Ok let’s take the Risk and leave the dep PR in. Else it is getting too complicated. :-(

I will review these as soon as possible, cause i want to look into it carefully and prevent potential errors...

PR #228, opened 2020-07-07
bumb version and add Changelog file

Apollon77:
BEFORE we merge this we need to finally decide about https://github.com/mcollina/node-coap/pull/211 and pot. add it ... or ot :-)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * (Apollon77) Fix: Add missing checks for Buffer length before reading from it
 * (Apollon77) Update all dependencies
 * (JsonMa) Fix: remove useless variables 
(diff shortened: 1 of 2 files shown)


JsonMa:
> BEFORE we merge this we need to finally decide about #211 and pot. add it ... or ot :-)

lgtm...We may merge #211 first.

Apollon77:
Just merge when you are ready and I Update here.

Apollon77:
@mcollina @JsonMa Changelog finalized ... this would be the 0.24.0 I think ...

JsonMa:
> @mcollina @JsonMa Changelog finalized ... this would be the 0.24.0 I think ...
Agree

Apollon77:
Thank you @mcollina

Issue #236, opened 2021-02-21
Problem with block-wise responses to multicast request


JKRhb:
Sending a GET request to an IPv6 multicast address currently seems to fail if the responses are transmitted using block-wise transfer. [RFC 7959](https://tools.ietf.org/html/rfc7959#section-2.8) states that

> the client retrieves any further blocks using unicast exchanges

after the initial multicast request. In my case, however, despite the fact that the `multicast` request option is set to `true`, the client does not switch to unicast requests after the first response (cf. the wireshark screenshot below). If clients are not responding in a block-wise manner but with "regular" reponses, multicast requests work as intended.

Is there another option that has to be enabled to support block-wise responses? Or is this actually something that has to be added to the package?

![grafik](https://user-images.githubusercontent.com/12641361/108634100-540d7e00-7478-11eb-9c5d-290552b3e4e7.png)

JKRhb:
I think I have narrowed down the issue to two problems:

1. In the `handle` function in `Agent.js` the IP address for further requests with the `Block2` option should be replaced with the response's source address (which can be obtained from the function's `rsinfo` parameter) if the initial request was sent to a multicast address
2. If I see it correctly, another problem arises from the fact that the same token is used for further requests after the initial one. Therefore, the client gets "confused" if more than one server uses it in its responses. 

I would try to provide a first version of a PR to fix these problems in the next few days.

PR #237, opened 2021-02-24
Fix handling of block2 reponses to multicast requests

JKRhb:
This PR offers a potential fix for #236. As mentioned in the issue (and suggested by [RFC 7959](https://tools.ietf.org/rfcmarkup/7959#section-2.8)), the agent now switches to unicast requests if a response containing a block2 option is given to a multicast request. 

This solves a problem that occured if two servers offering providing the same resource path are reachable over the same multicast address leading to confusion in the agent as the same block numbers are requested over and over again from the multicast address once a new reponse arrives.

This PR is still a bit WIP as the new `_convertMulticastToUnicastRequest` function could probably be refactored if there is a more elegant way to perform a deep copy of the initial request. I am also still working on corresponding unit tests. However, I performed a lot of manually testing/debugging which makes me confident that the solution should already work. In any case, feedback is highly appreciated :)

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -206,6 +206,11 @@ Agent.prototype._handle = function handle(msg, rsinfo, outSocket) {
     }
   }
   if (block2) {
+
+    if (req.multicast) {
+      req = this._convertMulticastToUnicastRequest(req, rsinfo)
+    } 
+
     // accumulate payload
     req._totalPayload = Buffer.concat([req._totalPayload, packet.payload])
 
@@ -384,8 +389,9 @@ Agent.prototype.request = function request(url) {
       if (req._packet.token && req._packet.token.length >= 4) {
         delete that._tkToReq[req._packet.token.readUInt32BE(0)]
       }
+      delete that._msgIdToReq[req._packet.messageId]
       that._msgInFlight--
-      if (that._msgInFlight === 0) {
+      if (that._msgInFlight === 0 && that._closing) {
         that._doClose()
       }
     }, multicastTimeout)
@@ -422,6 +428,16 @@ function urlPropertyToPacketOption(url, req, property, option, separator) {
         buf.write(part)
         return buf
       }))
+}
+
+Agent.prototype._convertMulticastToUnicastRequest = function (req, rsinfo) {  
+  var unicastReq = this.request(req.url)
+  unicastReq.url.host, unicastReq.sender._host = rsinfo.address.split('%')[0]
+  unicastReq.url.multicast = false;
+  req.listeners("response").forEach(listener => {
+    unicastReq.on("response", listener)
+  });
+  return unicastReq
 }
 
 module.exports = Agent
--- a/test/request.js
+++ b/test/request.js
@@ -1398,7 +1398,37 @@ describe('request', function() {
       }).end()
     })
 
-  })
+    it('should allow for block-wise transfer when using multicast', function (done) {
+      var payload = Buffer.alloc(1536)
+        , counter = 0  
+      
+      server = coap.createServer((req, res) => {
+        expect(req.url).to.eql("/hello")
+        res.end(payload)
+      })
+      server.listen(sock)
+      
+      var server2 = coap.createServer((req, res) => {
+        expect(req.url).to.eql("/hello")
+        res.end(payload)
+      })
+      server2.listen(sock)
 
+      var _req = request({
+        host: MULTICAST_ADDR,
+        port: port2,
+        pathname: "/hello",
+        confirmable: false,
+        multicast: true,
+      }).on('response', function (res) {
+        expect(res.payload.toString()).to.eql(payload.toString())
+        counter++
+        if (counter == 2) {
+          done()
+        }
+      }).end()
+    })
+
+  })
 
 })


JKRhb:
@mcollina @Apollon77 The PR still needs the test to be added but could you maybe already have a look on it? (I hope it is okay that I ping you this way.)

mcollina:
Works for me! Add the test! There is probably some maintainance to do on this repo as well, let me know if you'd like to be added here.

JKRhb:
> Works for me! Add the test! 

Great :) Will do!

> There is probably some maintainance to do on this repo as well, let me know if you'd like to be added here.

I'd be very glad!

JKRhb:
I added now a new test case and actually was able to find (and fix) a small bug. I think this PR should now be ready to be merged :)

lib/agent.js
```diff
@@ -424,4 +430,16 @@ function urlPropertyToPacketOption(url, req, property, option, separator) {
       }))
 }
 
+Agent.prototype._convertMulticastToUnicastRequest = function (req, rsinfo) {
+  var reqOpts = {"host": rsinfo.address.split('%')[0], "port": rsinfo.port}
+  var unicast_req = this.request(reqOpts)
+  unicast_req.url = req.url
+  unicast_req.url.host = unicast_req.sender._host
+  unicast_req.url.multicast = false;
+  unicast_req._events = req._events
```
mcollina, line 439: Is this needed? I think this might create significant problems as this is internal state of EventEmitter.

mcollina (changes requested):

lib/agent.js
JKRhb: You are right, this is rather suboptimal. The first three lines weren't actually needed and I removed them in the fixup commits I just added. I also changed the fourth line now to 

```js
unicast_req._events.response = req._events.response
```

so only the function references are copied. Maybe this will work better?

lib/agent.js
```diff
@@ -424,4 +430,12 @@ function urlPropertyToPacketOption(url, req, property, option, separator) {
       }))
 }
 
+Agent.prototype._convertMulticastToUnicastRequest = function (req, rsinfo) {
+  var reqOpts = {"host": rsinfo.address.split('%')[0], "port": rsinfo.port}
+  var unicast_req = this.request(reqOpts)
+  unicast_req._events.response = req._events.response
```
mcollina, line 436: Do not touch `_events`. Use public methods from `EventEmitter` to interact with it/

lib/agent.js
JKRhb: (I took the liberty of already squashing the commits.)

lib/agent.js
JKRhb: Oh, I see, thank you. I will fix this.

lib/agent.js
```diff
@@ -424,4 +430,14 @@ function urlPropertyToPacketOption(url, req, property, option, separator) {
       }))
 }
 
+Agent.prototype._convertMulticastToUnicastRequest = function (req, rsinfo) {
+  var reqOpts = {"host": rsinfo.address.split('%')[0], "port": rsinfo.port}
+  var unicast_req = this.request(reqOpts)
+  req.listeners("response").forEach(listener => {
+    unicast_req.on("response", listener)
+  });
```
JKRhb, line 438: Does this work? @mcollina

lib/agent.js
```diff
@@ -424,4 +430,14 @@ function urlPropertyToPacketOption(url, req, property, option, separator) {
...
+  unicast_req._packet.token = this._nextToken()
```
JKRhb, line 439: I just noticed that this line is unnecessary as well.

JKRhb: Fixed.

lib/agent.js
```diff
@@ -424,4 +430,14 @@ function urlPropertyToPacketOption(url, req, property, option, separator) {
       }))
 }
 
+Agent.prototype._convertMulticastToUnicastRequest = function (req, rsinfo) {  
+  var unicastReq = this.request(req.url)
+  unicastReq.url.host, unicastReq.sender._host = rsinfo.address.split('%')[0]
+  unicastReq.url.multicast = false;
```
JKRhb, line 436: I just noticed that only copying the address and port is not enough to let the unicast request behave exactly like the multicast one. This is my new attempt to perform something that should be equivalent of a deep copy of the used parameters. Testing this manually with a device that exposes a resource under a path already worked, I would also add a unit test for this.

I also changed the casing from snake_case to camelCase.

test/request.js
```diff
@@ -1398,7 +1398,37 @@ describe('request', function() {
...
+      var server2 = coap.createServer((req, res) => {
+        expect(req.url).to.eql("/hello")
+        res.end(payload)
+      })
+      server2.listen(sock)
+
+      var _req = request({
+        host: MULTICAST_ADDR,
+        port: port2,
+        pathname: "/hello",
```
JKRhb, line 1420: I extended now the existing test to check for the correct URI path. I think this should work now.

mcollina (approved):
lgtm

mcollina:
I think this is good to go!

JKRhb:
Great! :) Do you want to hit merge button?

PR #240, opened 2021-03-13
chore: move CI to Github Actions

### LINKED ISSUE #224: Switch testing to GitHub Actions

JKRhb:
This PR will resolve #224. Tests are now performed on Node versions 10.x, 12.x, 14.x, 15.x, and on multiple platforms (Ubuntu, Windows, Mac OS). One test on Mac OS, however, fails at the moment. This problem should either be fixed or Mac OS should be skipped for testing for now.

The CI pipeline now also generates a coverage report and supports `coveralls` which can be used for including a badge in the readme. I also thought about including a linting step using JSHint or ESLint but I guess this would require some configuration due to the used coding style.

### DIFF
--- a/.github/workflows/npm-test.yml
+++ b/.github/workflows/npm-test.yml
@@ -1,0 +1,38 @@
+name: Build Status
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        node-version: [10.x, 12.x, 14.x, 15.x]
+        os: [ubuntu-latest, windows-latest] # FIXME: Tests on macos-latest fail at the moment
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+
+    - name: Install dependencies
+      run: npm install
+
+    - name: Perform tests and generate coverage report
+      run: |
+        npm test
+        npm run coverage
+
+    - name: Coveralls
+      uses: coverallsapp/github-action@master
+      with:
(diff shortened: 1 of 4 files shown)


Apollon77:
LGTM

Apollon77 (approved):
... surely after deciding about the macos topic :-)

Apollon77:
@JKRhb Maybe add in the test a console log temporary to see if thata are not received or only partially or whatever happens

JKRhb:
> 
> 
> @JKRhb Maybe add in the test a console log temporary to see if thata are not received or only partially or whatever happens

The results of the test in question can be seen [here](https://github.com/JKRhb/node-coap/runs/2102671935?check_suite_focus=true). It seems to be an issue with shared sockets

test/share-socket.js
```diff
@@ -26,6 +26,7 @@ describe('share-socket', function() {
   })
 
   afterEach(function(done) {
+    this.skip()
     this.timeout(200)
     setTimeout(function() {
       server.close(done)
```
JKRhb, line 32: I actually narrowed down the problem now to this `afterEach` hook. If it is skipped the tests pass on Mac OS. Maybe increasing the timeouts could fix the problem?

JKRhb: Hmm. I did some more investigation and somehow the tests in `share-socket.js` fail randomly with the following error message:

` Caught exception: Error [ERR_SOCKET_DGRAM_NOT_RUNNING]: Not running`

Maybe this is a more general problem with UDP sockets in NodeJS under Mac OS?

JKRhb:
I tried out a lot of different parameters now and did not find a reliable way to get the tests to work on Mac OS. I would suggest limiting the CI to Ubuntu and Windows for now (which is an extension of the old testing capabilities) and opening an issue for the remaining tests on Mac OS. Maybe this is actually a more general problem with UDP sockets and NodeJS on Mac OS that has be solved somewhere else.

mcollina (approved):
lgtm

mcollina:
this needs to land first, it'd be active on master.

PR #242, opened 2021-03-17
fix: copy all listeners when using block2 multicast

JKRhb:
This PR fixes a small bug that was included in #237 by "copying" not only the listeners for `response` events but for all events registered to a multicast request when converting it to a unicast request. Copying only `reponse` listeners could lead, for example, to unhandled `error` events causing an application to crash.


mcollina (changes requested):
Can you add a test?

JKRhb:
> 
> 
> Can you add a test?

Done :)

mcollina (approved):
lgtm

PR #243, opened 2021-03-21
Fix three more problems with blockwise multicast responses

JKRhb:
I noticed three more (and hopefully last) problems with block-wise transfer and multicast requests which are addressed by this PR. 

First, I noticed that when performing a multicast request nodes could reply multiple times leading to more messages than required and, more importantly, duplicated payloads which could lead to unintended behavior. I fixed this problem by keeping track of the hostnames of the respondants that replied to the inital multicast request. If a server replies twice to a multicast, all reponses after the first one will be ignored. 

A second fix addresses the fact that the multicast timer from the initial request got reinitialized in the converted unicast request causing the client to abort the transfer once the time limit was reached. Using `clearTimeout()`, this timer now gets removed.

Third, new tokens are now assigned to the created unicast requests. Before, no new token was created which lead to issues when more than one server was responding to the multicast. A final minor change is the now correct assignment of the unicast IP address to the new request's `url.host` field (which should have little to no impact, though).

Because of the changes to the way "duplicated" IP addresses are handled, I had to adjust the two test cases I added in #237 and #242. Despite the fact that now only one server is left I think the core functionality is still tested as the single server is responding to the multicast in a blockwise manner despite switching from multicast to unicast.

Sorry that this is yet another PR on essentially the same issue. However, I think now I found a solution that should work, so hopefully no more PRs on this will be needed :)

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -97,6 +97,7 @@ Agent.prototype._init = function initSock(socket) {
 
   this._msgIdToReq = {}
   this._tkToReq = {}
+  this._tkToMulticastResAddr = {}
 
   this._lastToken = Math.floor(Math.random() * (maxToken - 1))
   this._lastMessageId = Math.floor(Math.random() * (maxMessageId - 1))
@@ -209,7 +210,10 @@ Agent.prototype._handle = function handle(msg, rsinfo, outSocket) {
 
     if (req.multicast) {
       req = this._convertMulticastToUnicastRequest(req, rsinfo)
-    } 
+      if (!req) {
+        return
+      }
+    }
 
     // accumulate payload
     req._totalPayload = Buffer.concat([req._totalPayload, packet.payload])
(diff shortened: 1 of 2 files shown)


Apollon77:
@JKRhb One question ... we had a strange issue with Shelly devices recently. They now also allow to configure a device to use unicast. The effeact of that was that as soon  as ONE device was changed to unicast in fact ALL messages came in with the IP of that unicast device - even if they were received via multicast.

Coud this be connected to your third part here?

JKRhb:
> @JKRhb One question ... we had a strange issue with Shelly devices recently. They now also allow to configure a device to use unicast. The effeact of that was that as soon as ONE device was changed to unicast in fact ALL messages came in with the IP of that unicast device - even if they were received via multicast.
> 
> Coud this be connected to your third part here?

I think that could actually be the case. IIRC in my case a mismatch of responses happened if more than one device replied to the same multicast because of the missing tokens. Could you maybe perform another test with my branch?

mcollina (approved):
lgtm

JKRhb:
> lgtm

Great, thank you :) I would add one more test and then I would merge the PR

PR #244, opened 2021-05-16
Add linter for StandardJS

JKRhb:
This PR adds a linter for Standard JS to both the CI pipeline and the `package.json` file.

It does not include any fixes yet as there are almost 4000 issues in the code but the linter already gives feedback to newly added code under the changes tab which should be handy. Follow-up PRs should address issues in the code one-by-one and gradually move the codebase towards StandardJS compliance.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,40 @@
+name: StandardJS Linter
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  lint:
+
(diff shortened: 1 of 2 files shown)


Apollon77:
@JKRhb Can you update the PR please and fix the linter warnings? ;-))

JKRhb:
> @JKRhb Can you update the PR please and fix the linter warnings? ;-))

I guess we could actually close this one and just proceed with #251 which includes this PR and goes beyond it :)

Apollon77:
agreed

PR #190, opened 2018-06-15
Feat/188

everhardt:
An attempt to implement https://github.com/mcollina/node-coap/issues/188. Also added a test and made a comment about it in README.md

PR #246, opened 2021-06-12
Add version info on type 1 block-wise transfers to README

JKRhb:
This PR updates the README in accordance to the newly added feature for type 1 block-wise transfer. I created a PR for this change as I wanted to be sure that this actually going to be a new *minor* release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,7 +320,7 @@ Since v0.7.0, this library supports blockwise transfers, you can trigger
 them by adding a `req.setOption('Block2', Buffer.of(0x2))` to the
 output of [request](#request).
 
-And since v?.?.?, this library supports rudimentry type 1 blockwise transfers, you can trigger
+And since v0.25.0, this library supports rudimentry type 1 blockwise transfers, you can trigger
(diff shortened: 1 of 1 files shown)


Apollon77:
Then lets release it :-)

JKRhb:
> Then lets release it :-)

Yay :) 

Speaking of which: Did you get the npm access by @mcollina, yet? :)

Apollon77:
Now that you ask... don't think so ... @mcollina  :-)

mcollina:
@Apollon77 I've added you.

Apollon77:
@mcollina Thank you for your trust!

@JKRhb So how to continue? Release what we have now or do the other stuff and do a big one? ;)

JKRhb:
> @JKRhb So how to continue? Release what we have now or do the other stuff and do a big one? ;)

Maybe we could already go for a release as the other PRs (except #190) are more style/refactoring related? :) After #251 is merged we could maybe publish the changes as a patch?

Apollon77:
Then lets do it that way ...

PR #248, opened 2021-06-15
feat: add more content-formats

JKRhb:
This PR adds more content-formats the library that are available by default. There are still codes that could be added, see [this IANA page](https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#content-formats) for the full list of available CoAP content-formats.

I am not sure about the current way the formats are registered. Maybe it makes more sense to define them as an array somewhere and then iterate over them at startup?

### DIFF
--- a/lib/option_converter.js
+++ b/lib/option_converter.js
@@ -112,13 +112,36 @@ var registerFormat = function(name, value) {
 
 module.exports.registerFormat = registerFormat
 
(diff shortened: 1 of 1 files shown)


Apollon77:
LGTM

JKRhb:
> LGTM

Nice, then I'm going to merge now :)

PR #250, opened 2021-06-15
Update copyright information

JKRhb:
This PR updates the copyright information to "2013-2021" and adds a license header to two newly added files that didn't have one before.

### DIFF
--- a/lib/segmentation.js
+++ b/lib/segmentation.js
@@ -1,3 +1,13 @@
+'use strict';
+
+/*
+ * Copyright (c) 2013-2021 node-coap contributors.
+ *
+ * node-coap is licensed under an MIT +no-false-attribs license.
+ * All rights not explicitly granted in the MIT license are reserved.
+ * See the included LICENSE file for more details.
+ */
+
 var generate           = require('coap-packet').generate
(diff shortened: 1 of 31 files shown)


JKRhb:
(@Apollon77 Let me know if I request you too often for reviews. If this should be the case I will simply open PRs and you can have a look into them if you find the time.)

lib/segmentation.js
```diff
@@ -1,3 +1,13 @@
+'use strict';
+
+/*
+ * Copyright (c) 2013-2021 node-coap contributors.
```
JKRhb, line 4: I was wondering if the newly added files should also have a copyright that begins in 2013? On the other hand it probably doesn't make much of a difference.

Apollon77: Formally a file copyright should start when file was added, but yes in the end ... no real dofference :-)

Apollon77:
@JKRhb Requesting review is best  ... so I get a mail and have a better triger and see when a PR is ready ...

JKRhb:
> @JKRhb Requesting review is best ... so I get a mail and have a better triger and see when a PR is ready ...

Okay, great :)

PR #251, opened 2021-06-16
Fix coding style with regard to StandardJS

### LINKED ISSUE #245: Add linter to CI and clean up code
If I am not mistaken the project is supposed to use StandardJS as its coding style. However, there are a lot of style issues and inconsistency at the moment. Enforcing a stricter coding style could be benefetial to make the project more easily to maintain.

Regarding StandardJS I opened a PR (#244) recently that introduces a Github action for linting and changes the code base accordingly. However, there are probably way too many changes in this PR to keep track of them which is why I opened the PR as a draft.

My suggestion would be to introduce only a very basic linter for now and fix only the most basic issues regarding StandardJS. After that the linting rules could be moved gradually towards StandardJS compliance. What do you think?

JKRhb:
This PR updates the codebase to StandardJS and fixes some minor issues. It is still a bit WIP as there are some minor issues left that need to be addressed.

This PR is supposed to be merged after (or as an alternative) to #244 and closes #245.


JKRhb:
@Apollon77 I managed to update the whole codebase to StandardJS now. I am not sure, however, if this PR is really reviewable :/

Apollon77:
I did a fly over - the hugh test coverage is key here I think and in this case testing still works which is a good sign.

I would go for it

JKRhb:
> I did a fly over - the hugh test coverage is key here I think and in this case testing still works which is a good sign.
> 
> I would go for it

Okay, great :) Then I guess we could close #255 and go for this PR instead. Maybe we could add another rule that overrides StandardJS' two spaces of indentation and replaces them with four. Even though I am personally okay with two spaces you are right that four spaces make lines easier to distinguish.

Apollon77:
Then lets do it that way and thank you for seeing my arguments :-)

JKRhb:
> Then lets do it that way and thank you for seeing my arguments :-)

Great, I will soon update this PR accordingly :)

JKRhb:
@Apollon77 I've applied the new identation now and rebased the PR against `master` :) I guess it should be able to be merged now after the release of `0.25.0`.

lib/block.js
```diff
@@ -6,74 +6,75 @@
...
-    sequenceNumber = sequenceNumber.sequenceNumber
-  }
+module.exports.generateBlockOption = function (sequenceNumber, moreBlocks, blockSize) {
+    if (typeof sequenceNumber === 'object') {
+        moreBlocks = sequenceNumber.moreBlocks
+        blockSize = sequenceNumber.blockSize
+        sequenceNumber = sequenceNumber.sequenceNumber
+    }
+
+    if (isNullOrUndefined(sequenceNumber) || isNullOrUndefined(blockSize) || isNullOrUndefined(sequenceNumber)) {
```
JKRhb, line 26: I just noticed that this line can be refactored like so:

```suggestion
    if (sequenceNumber == Null || blockSize == Null || sequenceNumber == Null {
```

This change would make the `isNullOrUndefined` function obsolete.

Apollon77: But the change could lead with some eslink checks to "== is bad because types can change blaaahh" warnings .. so we should suppress them? ;) but agreed that just because if this having a method makes not that much sense

JKRhb: The check for ´null` and `undefined` is actually the one expection where can use `==` instead of `===` according to their [rules](https://standardjs.com/rules.html) (see `eqeqeq`). So I think it should be safe to use it without suppressing them :)

I noticed, though, that I made three typos in the original suggestion (`Null` instead of `null`🤦‍♂️). Those are fixed now, however :)

Apollon77: Yes also wanted to mentionthe capital "N"s, but lets do it that way.

JKRhb:
I am going to rebase this branch with the new changes from the `master` branch in the upcoming days.

Apollon77:
Then rebase here and close #255 right?

JKRhb:
> Then rebase here and close #255 right?

Yeah, you are right :) I now managed to resolve all conflicts and all tests are still passing. Maybe you could do a last fly-over before we merge this thing? :)

JKRhb:
I added a couple of last minute changes in 28781b3dfbf5387a8b9200a6d04ec3092d19c3ee that should make control flow statements a bit more readable.

Apollon77:
Will check next days (on vacation right now, so time is rare)

JKRhb:
> Will check next days (on vacation right now, so time is rare)

No problem, take your time :) I just wanted to get this off my list. Hope you are having a good time :)

PR #254, opened 2021-06-22
style: remove obsolete semicolons

JKRhb:
This PR removes obsolete semicolons from the source code and introduces linting with ESLint. It is a softer approach than #251. After the biggest issues with the code have been resolved the StandardJS linter can be added as a last step.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,0 +1,40 @@
+name: StandardJS Linter
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  lint:
+
+    runs-on: ${{ matrix.os }}
+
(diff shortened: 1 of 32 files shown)


JKRhb:
@Apollon77 I decided now to tackle the style issue one step at a time (sorry for my obsession with this). Maybe you could have a quick look at this PR and hit the merge button? :)

Apollon77:
you're welcome :-) If there is a style it should be consistent (I personally like the "semicolon at the end" more.but other story ;-))

JKRhb:
> you're welcome :-) If there is a style it should be consistent (I personally like the "semicolon at the end" more.but other story ;-))

Thank you! :) Yeah, I guess the use or non-use of semicolons is quite disputed :D But you are right, consistency is key :)

PR #255, opened 2021-06-22
style: enforce indentation using two spaces

JKRhb:
This PR fixes the identation used in the repository and adds an eslint rule.

### DIFF
--- a/examples/blockwise.js
+++ b/examples/blockwise.js
@@ -1,10 +1,10 @@
 const coap        = require('../') // or coap
-		, url 				= require('url')
+  , url 				= require('url')
 
 coap.createServer(function(req, res) {
-	var path = url.parse(req.url)
-	var time = parseInt(path.search.split('=')[1])
-	var pathname = path.pathname.split('/')[1]
+  var path = url.parse(req.url)
+  var time = parseInt(path.search.split('=')[1])
+  var pathname = path.pathname.split('/')[1]
 
   res.end(new Array(time+1).join(pathname+' '))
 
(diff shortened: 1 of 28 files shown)


JKRhb:
@Apollon77 Ehem, here is another one 😅 Should I just go forward with changes like these? I would transform the codebase gradually towards StandardJS compliance and then finally merge #251.

Apollon77:
Puuhh ... Honestly this is the main part I do not like on TandardJS ... In my eyes 2 spaces is much less readable then 4 spaces ... and I use 4 spaces in all my other projects and we had such discussions several times and readability was always the main argument.
How you see it?
In this case the standard "do not support" me and so I don't care about standards in such cases ;-))

Apollon77:
@mcollina @jkivilin maybe also opinions out there ?!

Apollon77:
@JKRhb seems noone else have an opinion :-) can we stay with the 4 spaces? ;-)

JKRhb:
> @JKRhb seems noone else have an opinion :-) can we stay with the 4 spaces? ;-)

Sure :) I will adjust the PR accordingly :) Regarding #251 I think it might be useful in general to create our own coding style as StandardJS is too far away from the current code base to keep the changes from being too drastic. I will maybe add a (not too extensive) list of possible ESlint rules to #251 where we can discuss if we want to add them.

I think the most useful change will be converting all those `var`s to `const` and `let` which will help with finding bugs. Apart of from this there are probably things that deserve a higher priority than coding style in the package – sorry again for being a bit obsessed with this topic ;)

JKRhb:
Closed in favor of #255. CC @Apollon77

PR #256, opened 2021-06-26
feat: add `clientIdentifier` option to `createServer`

JKRhb:
This PR is a follow-up to #190. It fixes a couple of issues that were present on that branch, cleans up the commit history, and removes the added `package.json` file.

### DIFF
--- a/test/server.js
+++ b/test/server.js
@@ -1265,7 +1265,7 @@ describe('server block cache', function() {
 
 })
 
-describe('server LRU', function() {
+describe('Client Identifier', function() {
   var server
       , port
       , clientPort
--- a/test/server.js
+++ b/test/server.js
@@ -1264,3 +1264,103 @@ describe('server block cache', function() {
   })
 
 })
+
+describe('server LRU', function() {
+  var server
+      , port
+      , clientPort
+      , client
+      , clock
+
+  const packet = function(key) {
+    return {
+      confirmable: true
+      , messageId: 4242
+      , token: new Buffer(5)
+      , options: [
+        {name: 2109, value: Buffer.from(key) }
+      ]
+    }
+  }
+
+  beforeEach(function (done) {
+    clock = sinon.useFakeTimers()
+    port = nextPort()
+
+    server = coap.createServer({
+      clientIdentifier: (request) => {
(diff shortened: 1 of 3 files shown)


test/server.js
```diff
@@ -1263,4 +1263,104 @@ describe('server block cache', function() {
     })
   })
 
-})
\ No newline at end of file
+})
+
+describe('server LRU', function() {
```
JKRhb, line 1268: I am not sure about the naming for these test cases. In [line 1152](https://github.com/mcollina/node-coap/pull/256/files#diff-d14cbfb50ed1cad7db5f4fef6a6076961b7cc9be980a3be06a70998f0eb8ebceR1152), there is already a `server LRU` present. Maybe the tests can be merged?

JKRhb:
@Apollon77 Could you maybe have a quick look on this one? :) Maybe it could be integrated into the 0.25.0 release.

Apollon77:
Sure, I ran a bit outt of sync with some tasks ... will check tonight

JKRhb:
> will check tonight

Great, thank you!

test/server.js
Apollon77: Merging is one option, having the name state an "whats the test focus on" infgo is also an idea. tests should not become too complex imo

Apollon77:
@JKRhb Basically ok, I left you one comment for the test question ... what you think?

test/server.js
JKRhb: > , having the name state an "whats the test focus on" infgo is also an idea. tests should not become too complex imo

Yeah, you are right. What do you think about this one?

```suggestion
describe('Client Identifier', function() {
```

 (Not sure why I haven't been thinking about this name earlier 😅)

JKRhb:
> @JKRhb Basically ok, I left you one comment for the test question ... what you think?

Thank you :) I replied with a naming suggestion :)

Apollon77:
accepted ... lets formally have test running and then have fun merging it :-)

JKRhb:
> accepted ... lets formally have test running and then have fun merging it :-)

Awesome :) Thanks for the review!

Apollon77:
> Thanks for the review!

yw

PR #264, opened 2021-09-13
feat: add support for FETCH, PATCH, and iPATCH

JKRhb:
This PR adds support for the FETCH, PATCH, and iPATCH CoAP methods, as specified in [RFC 8132](https://datatracker.ietf.org/doc/html/rfc8132). I still need to figure out what is actually needed to fully support these methods (besides the status codes), so this PR is for now only opened as a draft.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,8 @@ This library follows:
 * [RFC 7641](https://datatracker.ietf.org/doc/html/rfc7641)
   for the observe specification,
 * [RFC 7959](https://datatracker.ietf.org/doc/html/rfc7959) for
-  the blockwise specification.
+  the blockwise specification,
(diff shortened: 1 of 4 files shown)


JKRhb:
@Apollon77 It would be awesome if you could have a look on this PR :) For me it seems that there is actually no more logic necessary to support the new methods but maybe you could also have a look into RFC 8132?

Apollon77:
I try to check next days ... seems "a lot" that you managed to do :-)

JKRhb:
@Apollon77 I think this PR is now the only one that is left :) Would be awesome if you could have a look on it and proceed with merging if it is okay. I guess after that we can create a new release :)

JKRhb:
@Apollon77 As this PR does not introduce the most complex new feature I would go ahead and merge it :) Let me know if you notice anything that should be changed then we could open a follow-up PR :)

Apollon77:
LGTM

PR #266, opened 2021-09-17
fix(agent): reset _lastMessageId to 0

JKRhb:
This is a small fix which allows message IDs to be set to 0 if the maximum message ID value has been reached. This makes the implementation a bit more consistent as the initial message ID can actually be 0.

As a bug in `coap-packet` has been fixed that caused errors when the ID was 0 this change should be fine to add now.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -332,7 +332,7 @@ class Agent extends events.EventEmitter {
 
     _nextMessageId () {
         if (++this._lastMessageId === maxMessageId) {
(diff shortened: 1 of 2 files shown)


JKRhb:
@Apollon77 Maybe we could publish a new version after #264 and this PR have been merged? :)

JKRhb:
@Apollon77 As this is very trivial fix I would simply go ahead and merge this PR :)

Apollon77:
LGTM

PR #267, opened 2021-09-17
refactor: replace deprecated URL.parse

JKRhb:
This PR replaces `URL.parse` which is deprecated according to eslint. It fixes #267.


JKRhb:
@Apollon77 Same here. I think the changes shouldn't be too critical so I would proceed with merging :)

Apollon77:
LGTM

PR #269, opened 2021-09-21
fix: add support for `text/plain; charset=utf-8`

### LINKED ISSUE #33: Format 0 must be utf-8
Format 0 is defined as `text/plain; charset=utf-8`. By definition, text/\* defaults to US-ASCII and causes encoding problems if a Unicode-encoded document is served without this charset parameter - e.g. in CoAP-to-HTTP translators.

Instead, the module serves option 0 as merely `text/plain` i.e. US-ASCII.

JKRhb:
This PR fixes #33. It adds support for the Content-Format `text/plain; charset=utf-8` as an input parameter. The default output Content-Format will now be changed to `text/plain; charset=utf-8` so this might actually be a breaking change. I hope, however, that other implementations will be able to handle this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -501,6 +501,7 @@ registry](https://www.iana.org/assignments/core-parameters/core-parameters.xhtml
 These are the defaults formats:
 ```js
 registerFormat('text/plain', 0)
+registerFormat('text/plain; charset=utf-8', 0)
 registerFormat('application/link-format', 40)
 registerFormat('application/xml', 41)
 registerFormat('application/octet-stream', 42)
--- a/lib/option_converter.js
+++ b/lib/option_converter.js
@@ -118,6 +118,7 @@ module.exports.registerFormat = registerFormat
 // See https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#content-formats
 // for a list of all registered content-formats
 registerFormat('text/plain', 0)
+registerFormat('text/plain; charset=utf-8', 0)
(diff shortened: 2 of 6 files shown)


JKRhb:
We might have to find a better way for dealing with content-formats which contain additional parameters in the future. At the moment, only the part before the first `;` is being used. This approach, however, is not usable with new content-formats like `application/cose; cose-type="cose-encrypt"` or `application/cose; cose-type="cose-mac"` where the parameters actually contain important information.

JKRhb:
@awwright: Would you still be interested in this issue?

awwright:
I recall the reason I didn't was because it could break backwards compatibility in applications that use merely `text/plain` to refer to format 0. I'm not sure how much of a problem this actually is or a good way to fix it.

JKRhb:
> I recall the reason I didn't was because it could break backwards compatibility in applications that use merely `text/plain` to refer to format 0. I'm not sure how much of a problem this actually is or a good way to fix it.

Hmm, yeah, I am not so sure either. Maybe a more complex process (i. e. actually parsing the parameters) is needed to distinguish different content-formats, at least when they are given as an input. In any case, considering the potential breaking change it might be better to defer this PR to version 1.0.0.

JKRhb:
I now opened and merged #282 which deals (partly) with the parameters issue. I am closing this PR for now but Content-Format 0 should probably be revisited in the future.

PR #271, opened 2021-09-24
fix(server): allow only one server instance for each port number

### LINKED ISSUE #268: Multiple server instances can listen to the same port
I'm using coap 0.25.0 and trying to listen to the same port using two separate Coap server instances and I though it would have thrown error when attempting to listen again on the same port using `listen(port, address, done)`. It comes out I am wrong, but is it a bug or what?

Here's a minimal sample of what I mean:
```
const coap = require("coap");

const port = 5683;
const address = "localhost";

let server = coap.createServer();
let server2 = coap.createServer();

server.once("error", (err) => {
    throw new Error(err);
});

server2.once("error", (err) => {
    throw new Error(err);
});

server.listen(port, address, (res) => {
    console.log(`error is ${res}`);
});

server2.listen(port, address, (res) => {
    console.log(`error is ${res}`);
});
```

Both the `done` callbacks correctly return and `err` is undefined. Shouldn't the second call return an error? Or I just have to assume that multiple servers can listen to connections on the same port?

JKRhb:
This PR would fix #268. I am not sure, however, if this change is really necessary/desirable. I am looking forward to any comments on this :) 

CC @fillobotto @relu91

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -243,6 +243,10 @@ class CoAPServer extends events.EventEmitter {
 
         if (!this._options.type) this._options.type = 'udp4'
 
+        if (this._options.reuseAddr == null) {
+            this._options.reuseAddr = true
+        }
+
         if (port instanceof events.EventEmitter) {
             this._sock = port
             if (done) setImmediate(done)
@@ -250,7 +254,7 @@ class CoAPServer extends events.EventEmitter {
             this._internal_socket = true
             this._sock = dgram.createSocket({
                 type: this._options.type,
-                reuseAddr: true
+                reuseAddr: this._options.reuseAddr
             })
 
             this._sock.bind(port, address || null, function () {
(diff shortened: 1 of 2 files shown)


relu91 (changes requested):
Can't we expose this option somewhere? so that we can keep the current behavior but override it when it is needed?

JKRhb:
> Can't we expose this option somewhere? so that we can keep the current behavior but override it when it is needed?

Oh, that is a very good idea! I will add it to the PR :)

JKRhb:
@relu91 I updated the PR. Would be awesome if you could have a quick look then I'll guess this can be merged :)

relu91 (approved):
Looks good! Just remember to update also #278 when this lands 😄

JKRhb:
> Looks good! Just remember to update also #278 when this lands 😄

Will do, thanks for the reminder :D

lib/server.js
```diff
@@ -243,14 +243,18 @@ class CoAPServer extends events.EventEmitter {
 
         if (!this._options.type) this._options.type = 'udp4'
 
+        if (this._options.reuseAddr == null) {
```
Apollon77, line 246: I would have not done the == null check that way (more like !== false ... is true :-) and only false is false ... would be my comment here :-)

Apollon77:
see one comment, but just nitpick

lib/server.js
JKRhb: Oh, very good point! I will fix it in a follow-up PR :) Thank you for going through all of these PRs by the way! :)

Apollon77: Was shocked by the number :-) I'm not commenting all ... just some :-)

PR #272, opened 2021-09-24
refactor(agent): remove obsolete DNS bug check

JKRhb:
This PR removes code that checks for node version 4 which is EOL for a couple of years.


Apollon77:
LGTM

PR #273, opened 2021-09-25
refactor: replace prototypes with classes 

JKRhb:
This PR refactors the pre ECMAScript 2015 prototype approach used in a number of files and introduces classes instead. This should make the code a lot easier to read and also improve the project's maintainability

### DIFF
--- a/index.js
+++ b/index.js
@@ -40,7 +40,9 @@ module.exports.request = function (url) {
     return agent.request(url)
 }
 
-module.exports.createServer = Server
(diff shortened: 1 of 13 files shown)


JKRhb:
@Apollon77 I would simply go ahead and merge this PR as it only introduces some syntactic sugar – I hope that is okay for you. Otherwise let me know and we could revert this change :)

Apollon77:
All good, do not let yourself blocked by me, many other duties and not enough time. I put on my review list and review later

PR #274, opened 2021-09-25
refactor: replace `that` with `this` by using arrow functions

JKRhb:
This PR removes the

```js
const that = this
```

statements from the codebase and replaces occurences of `that` with `this` by using arrow functions which allow for calling `this` directly. This should make the code easier to read.


Apollon77:
thank you :-)) great improvement

PR #278, opened 2021-09-27
feat: add typescript declarations

JKRhb:
This PR adds a couple of declaration files to the project to enable typescript support. This might also be a first step for potentially porting the project to typescript in the future.

I added a first test to ensure that the type declarations work. There probably is some room for improvement, though. @relu91: Maybe you could also have a look on this PR? :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,6 @@ lib-cov
 
 #local builds
 *.tgz
+
+# typescript output
+dist
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,76 @@
+import Agent from './lib/agent'
+import Server from './lib/server'
+import IncomingMessage from './lib/incoming_message'
+import OutgoingMessage from './lib/outgoing_message'
+import { CoapMethod, OptionName } from 'coap-packet'
+import { Socket } from 'dgram'
+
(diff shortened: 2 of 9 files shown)


relu91 (commented):
First, thank you for asking to review this 😃 . It would be awesome to have typings for node-coap! Looking at your PR my feedback is to remove the `tsconfig.json` file inside ts and just use the root one. I don't see the benefits of having the second `tsconfing` inside the TS folder. Another point is that when you test ts you are not actually running the code. As suggested [here](https://github.com/DefinitelyTyped/DefinitelyTyped#adding-tests-to-a-new-package) it would be beneficial to emit js files and check if everything is running properly. What about changing tstests to `"tstest": "tsc && node out/test.js"` ?

Just my 2 cents

JKRhb:
Thank you for your feedback @relu91! :) Those are very good points and I will adjust the PR accordingly :)

JKRhb:
@relu91 Thanks again for your feedback! I think this PR should probably be ready now. I already tried to integrate the new types into node-wot locally and this should work as well :) So I guess once a new version is released we can actually switch to typed definitions in CoAP.

relu91 (approved):

PR #284, opened 2021-09-28
chore: ignore auxiliary files when publishing npm package

JKRhb:
I noticed that there a couple of configuration files as well as the coverage output included in published packages. This PR adjusts the `.npmignore` file accordingly as it overwrites `.gitignore` files when it exists.


Apollon77:
Thank you! Good thing

PR #285, opened 2021-09-28
feat: add partial typescript migration

JKRhb:
This PR attempts to add a first version of "native" typescript support to the project. It moves the contents of `index.js` to an `index.ts` file, includes typescript compilation into the npm scripts, and creates the needed `d.ts` files now automatically based on JSDoc comments. This way it should be easier to maintain a typed codebase without manually comparing the declaration files with the source files. I hope that this approach makes sense.


JKRhb:
@relu91 May I request your feedback once more on this topic? :)

JKRhb:
I just had another idea: Maybe replacing the declaration files in with JSDoc comments might already be enough for a more dynamic approach to typing. I will try that out in another PR.

JKRhb:
Closed in favor of #289.

PR #289, opened 2021-09-30
refactor: replace declaration files with JSDoc comments

JKRhb:
This a more elegant alternative to #285.


relu91:
I think the JSDoc approach is more maintainable since it put the type definition closer to the code. Sadly, I did not have the experience with porting a JS project to TS. I would say that this is the right direction though.

JKRhb:
> I think the JSDoc approach is more maintainable since it put the type definition closer to the code. Sadly, I did not have the experience with porting a JS project to TS. I would say that this is the right direction though.

Awesome, thanks for the feedback! :) I think I actually found a setup that offers the best of both worlds, keeping the project structure mostly as it is while providing strict(er) type checking based on the JSDoc comments. I need to add a couple of commits then this PR will be able to be merged :)

JKRhb:
With the exception of some cleanup this PR should now be finally finished :) It should provide typing support for the whole project now. Unfortunately, this isn't possible without having some kind of compilation step as JSDoc comments are only parsed if depending Typescript have a value > 0 for `maxNodeModuleJsDepth` which should not be the case too often. I therefore added a `.ts` file back and let `tsc` compile the whole project before testing and publishing.

In the process of working on this PR I noticed that there is an incorrect type definition for the `bl` module. I opened https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56130 to this fix issue. Once it is merged, this PR will be ready to be merged as well.

PR #293, opened 2021-10-03
refactor: migrate project to Typescript

JKRhb:
This PR is a potential follow-up to #289 migrating the whole project to Typescript.


Apollon77:
wow, now you go forward :-))

invaderb:
This is awesome! great job so far :)

JKRhb:
@invaderb @Apollon77 Thank you for your feedback! :) I now migrated most of the code, there are still some issues, though, which I have to investigate. I guess the PR should be ready for review later this week :)

JKRhb:
After some intense ~~procrastination~~ debugging, I now managed to finish this PR and move the whole codebase to Typescript :) Feedback is highly appreciated!

Both tests and the linter currently fail here on GitHub because https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56130 hasn't been merged yet but once that is done it should be possible to merge this PR as well. I guess #289 has become obsolete now or do you think it might make sense to have some kind of intermediary step between the old state of the project and this PR?

To better keep track of the changes I'll might create a couple of smaller PRs from this one that introduce certain aspects (like the null checks or the use of Maps instead of Objects for caching) first and then finally merge the rest of this PR.

index.ts
```diff
@@ -0,0 +1,190 @@
...
+        agent = exports.globalAgentIPv6
+    } else {
+        agent = exports.globalAgent
+    }
+    return agent.request(requestParams)
+}
+
+export type CoapOptions = Partial<Record<OptionName, OptionValue>>
+
+export interface CoapPacket extends Packet {
```
invaderb, line 94: Can you move these interfaces to their own files in a domain/models folder and import them it would clean up this file quite a bit?

JKRhb: Thank you, that is a very good point! I will soon push a commit addressing your suggestion :)

JKRhb: Good catch! Should be fixed now :)

JKRhb:
After the merge of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56130, the tests are now finally passing :)

Apollon77:
seems a rebase is needed @JKRhb

JKRhb:
> seems a rebase is needed @JKRhb

Done :)

PR #295, opened 2021-10-06
refactor: use Maps instead of Objects for caching

JKRhb:
This PR introduces some small refactoring in preparation for #293.


Apollon77:
Good idea, Like

PR #296, opened 2021-10-06
refactor: unify block1 and block2 terminology

JKRhb:
So far both the data structures as well as the terminology in block1 and block2 block state options differed. In preparation for more consistent typing this PR applies some more cleanup to the project.

### DIFF
--- a/lib/block.js
+++ b/lib/block.js
@@ -7,35 +7,39 @@
  */
 
 const TwoPowTwenty = 1048575
+
 /**
  *
- * @param {Number} sequenceNumber The block sequence number.
- * @param {Boolean} moreBlocks If there are more blocks to follow
- * @param {Number} blockSize
+ * @param numOrBlockState The block sequence number or a block state object.
+ * @param more Can indicate if more blocks are to follow.
+ * @param size The block size.
  */
-module.exports.generateBlockOption = (sequenceNumber, moreBlocks, blockSize) => {
-    if (typeof sequenceNumber === 'object') {
-        moreBlocks = sequenceNumber.moreBlocks
-        blockSize = sequenceNumber.blockSize
-        sequenceNumber = sequenceNumber.sequenceNumber
+module.exports.generateBlockOption = (numOrBlockState, more, size) => {
+    let num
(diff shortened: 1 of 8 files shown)


lib/block.js
```diff
@@ -7,35 +7,39 @@
...
 
 const TwoPowTwenty = 1048575
+
 /**
  *
- * @param {Number} sequenceNumber The block sequence number.
- * @param {Boolean} moreBlocks If there are more blocks to follow
- * @param {Number} blockSize
+ * @param numOrBlockState The block sequence number or a block state object.
+ * @param more Can indicate if more blocks are to follow.
```
Apollon77, line 14: "if" somehow indicated that it is a boolean ... but in fact it is a nunber?

lib/block.js
```diff
@@ -7,35 +7,39 @@
...
+    if (typeof numOrBlockState === 'object') {
+        num = numOrBlockState.num
+        more = numOrBlockState.more
+        size = numOrBlockState.size
+    } else {
+        num = numOrBlockState
     }
 
-    if (sequenceNumber == null || blockSize == null || sequenceNumber == null) {
+    if (num == null || more == null || size == null) {
```
Apollon77, line 27: if you want to make sure these are numbers maybe check that instead of == null

lib/block.js
Apollon77: and if number with only 0/1 why not boolean?

JKRhb: I think I changed it to a boolean first (there was some inconsistency as the block1 implementation used numbers, if I recall correctly) but then used 0/1 instead as the `more` flag can have different meanings depending on the context (see https://datatracker.ietf.org/doc/html/rfc7959#section-2.2). Therefore the "Can indicate". But I guess we could also use boolean values here.

lib/block.js
JKRhb: Oh, good point!

lib/block.js
Apollon77: Ok if specs have it as number we better stay near to the specs so maybe just change the text maybe to indicate that

PR #297, opened 2021-10-07
refactor: add strict nullchecks

JKRhb:
Another rather trivial PR adding explicit checks for `null` or `undefined`.

### DIFF
--- a/examples/json.js
+++ b/examples/json.js
@@ -25,7 +25,7 @@ coap.createServer((req, res) => {
             }
 
             res.pipe(bl((err, data) => {
-                if (err) {
+                if (err != null) {
                     process.exit(1)
                 } else {
                     const json = JSON.parse(data)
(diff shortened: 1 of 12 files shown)


Apollon77:
honestly I do not like "type-unsafe" checks with == or != ... They are not rteally better then the "falsy" checks before ...  (in my eyes) but all good we can leave it that way too ...

JKRhb:
Yeah, it adds some noise that isn't really necessary :/ The reason for PRs like these were the fact that I activated the option for strict null checks in #293 which complained about checks for falsy values. Maybe we can discuss over there if this is actually necessary. So far I thought the stricter the checks the better but in most cases it probably doesn't bring that much value.

Apollon77:
I personally use "falsy" checks if I know that it can not be false and want to do tese checks ... or I really check for types or such depending on what I want/need... but as said ... personal flavour

PR #300, opened 2021-10-07
refactor: remove `simplifyPacketForPrint` function

JKRhb:
This PR removes the `simplifyPacketForPrint` function from the helpers file. I noticed that this function isn't actually being used and as it causes some trouble when introducing type checking I wanted to propose removing it. I think it was introduced by @Jamezo97 and @invaderb, would this be okay for you? :)

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -176,20 +176,6 @@ module.exports.removeOption = (options, name) => {
     return false
 }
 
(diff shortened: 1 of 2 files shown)


JKRhb:
@Jamezo97 @invaderb I would merge this for now, let me know if there might still be a purpose for this function, then we could move it back in :)

PR #315, opened 2021-11-05
update deps

Apollon77:


Apollon77:
Hm ... should we kick nodejs 10? at least for testing?

JKRhb:
> Hm ... should we kick nodejs 10? at least for testing?

Oh, yeah, after some quick research I think upgrading the node version for the linter workflow could fix the CI pipeline :)

Apollon77:
aahh yes the linting was still done in nodejs 10. changed

Apollon77:
wtf ... https://github.com/mcollina/node-coap/runs/4122430279?check_suite_focus=true#step:5:4

JKRhb:
Hmm, maybe the error is caused by the caching step?

Apollon77:
Caching step? Could be ... lets see

JKRhb:
Hmm, my other theory would be that it is a node 16 problem. The installation fails under both Ubuntu and Windows using that node version, the others seem to be fine.

Apollon77:
what we should do?

Apollon77:
PS: (because just noticed it) ... should we publish the "test" dir in npm too?

JKRhb:
> PS: (because just noticed it) ... do we need to publish the "test" dir in npm too? (in fact we did)

Oh, sorry, that is a very good point! I will change that for the typescript version. Or can we publish a 0.26.1 from a different branch?

JKRhb:
> what we should do?

From what I've seen doing another quick search this problem might actually be related to the caching. Maybe we could try to comment that step out

Apollon77:
https://github.com/actions/cache/blob/main/examples.md#using-multiple-systems-and-npm-config

Apollon77:
But yes I will comment out

JKRhb:
Okay, my cache theory is disproven :D

Apollon77:
Ok, reson seems to be the eslint update and other packages ...

JKRhb:
Maybe downgrading eslint to 7.x.x (for now) could solve the problem?

Apollon77:
just did that :-)

Apollon77:
This was it

Apollon77:
happy review ;-)

JKRhb (approved):
LGTM ;)

PR #316, opened 2021-11-05
chore: ignore test files in npm package

JKRhb:
This PR is a potential fix for the problem that tests files end up on npm. The change to the `.npmignore` file should prevent that in the future.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -9,3 +9,5 @@ coverage
 .github
 .mocharc.json
 .nycrc.json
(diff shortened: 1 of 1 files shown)


Apollon77:
BTW: I also had that error in a local test https://github.com/mcollina/node-coap/runs/4122788149?check_suite_focus=true#step:6:578

Apollon77:
Restart tests most enough

PR #317, opened 2021-11-16
feat(agent): add `close` method

### LINKED ISSUE #312: Add `close()` method to `Agent`
The `Agent` class currently has no proper API for closing it externally. Adding a `close()` method might be good addition to be able to do so.

JKRhb:
This PR adds a `close` method to the `Agent` class that aborts all pending requests and resolves #312. To do so, a small bug is fixed in the `abort` method where the `_msgInFlight` hasn't been decremented when a request is cancelled.

An additional test is added to assert that the new method works as expected.

### DIFF
--- a/lib/agent.ts
+++ b/lib/agent.ts
@@ -103,6 +103,16 @@ class Agent extends EventEmitter {
         this._requests = 0
     }
 
+    close (done?: (err?: Error) => void): this {
+        for (const req of this._msgIdToReq.values()) {
+            this.abort(req)
+        }
+        if (done != null) {
+            setImmediate(done)
+        }
+        return this
+    }
+
     _cleanUp (): void {
         if (--this._requests !== 0) {
             return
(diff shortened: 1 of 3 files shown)


JKRhb:
Ping @relu91 :)

relu91 (approved):
I'm not coap expert but looking at the logic seems good. This should allow freeing all resources without waiting for responses :)

JKRhb:
> I'm not coap expert but looking at the logic seems good. This should allow freeing all resources without waiting for responses :)

Thank you for the feedback! :) I added a last couple of changes to the method so that you can both provide an optional callback and listen for `close` events. This way the API is more similar to the one the CoAP server provides. I think if @Apollon77 could also have a look on it then the PR should be able to be merged :)

Apollon77:
BTW: LGTM

JKRhb:
> Add that close before 1.0 or after? ;-)

Hmm, good question ;) I think we could already integrate it in 1.0 if this would be fine for you, but we could also put it into a separate minor (or patch?) release if you prefer.

Apollon77:
:-)

PR #322, opened 2021-12-01
fix(index.ts): fix issues with exported types

JKRhb:
This PR makes some small adjustments to `index.ts` where a couple of definitions weren't exported correctly. I think we could publish a (patch) version `1.0.1` after the merge of this PR.


Apollon77:
@JKRhb same process? You prepare the "version commit" then tag and I do npm?

PR #329, opened 2022-03-16
Improve close logic and fix corner cases

relu91:
Hello, this is my first PR to this repository let me know if I have to change the commit style or if I skipped some steps. Regarding the PR I'm coming from https://github.com/eclipse/thingweb.node-wot/issues/637 where we found that node hanged after using coap client.  I've dug into the code and I found out that the problem arises when you close an `agent` when there are no pending requests running. Basically if you take a look at the current implementation code:
```ts
 close (done?: (err?: Error) => void): this {
        for (const req of this._msgIdToReq.values()) {
            this.abort(req) // <-- never called if this._msgIdToReq.size === 0
        }
        if (done != null) {
            setImmediate(done)
        }
        return this
    }
```
The abort function is never called which in the end has the effect of not properly closing the socket. Net result? node hangs and does not exit. 

This PR fix this shortcoming and a couple of other details:
1. `done` function is only called after the socket is really closed
2. It adds a listener for errors in a test (on my machine whiteout that listener the test suite does not continue)

**Warning** after a clean clone on windows with node 14.17.0 I have a couple of tests that are still failing, this PR does not fix those.

### DIFF
--- a/lib/agent.ts
+++ b/lib/agent.ts
@@ -104,11 +104,16 @@ class Agent extends EventEmitter {
     }
 
     close (done?: (err?: Error) => void): this {
+        if (this._msgIdToReq.size === 0 && this._msgInFlight === 0) {
+            // No requests in flight, close immediately
+            this._doClose(done)
+            return this
+        }
+
+        done = done ?? (() => {})
+        this.once('close', done)
         for (const req of this._msgIdToReq.values()) {
             this.abort(req)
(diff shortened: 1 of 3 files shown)


JKRhb (approved):
Thank you for dealing with this issue, @relu91! 

@Apollon77: Could you release a new version after this PR? :)

PR #330, opened 2022-03-23
fix: fix problem with (boolean) observe parameters

JKRhb:
This PR (probably) fixes an issue raised in https://github.com/iobroker-community-adapters/ioBroker.philips-air/issues/15 (I think). It seems as if https://github.com/mcollina/node-coap/pull/309 broke handling of the observe option as the value provided by the user is converted into a numeric one while the internal check expects a boolean. I am not entirely sure, yet, if this is the actual fix for the problem but it is probably a step into the right direction.

### DIFF
--- a/models/models.ts
+++ b/models/models.ts
@@ -87,7 +87,7 @@ export interface CoapRequestParams {
     port?: number
     method?: CoapMethod
     confirmable?: boolean
-    observe?: 0 | 1 | boolean
+    observe?: 0 | 1 | boolean | string
     pathname?: string
     query?: string
     options?: Partial<Record<OptionName, OptionValue>>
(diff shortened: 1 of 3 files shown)


models/models.ts
```diff
@@ -87,7 +87,7 @@ export interface CoapRequestParams {
     port?: number
     method?: CoapMethod
     confirmable?: boolean
-    observe?: 0 | 1 | boolean
```
relu91, line 90: I was just passing by but why do you want to disallow people to use `false`? sorry for bothering I'm just curious. 😃

JKRhb: Good point actually ;) I think the rationale behind it was that a boolean value of `false` does not have any effect at the moment. It could be mapped to a value of 1 (i. e. unobserve) but that might constitute a breaking change.

In any case, I changed the type back to `boolean`, realizing that it might otherwise introduce other incompatabilities with node-wot.

PR #333, opened 2022-03-31
fix: move bl and lru-cache types to dependencies

JKRhb:
This PR fixes a typescript related issue raised in https://github.com/eclipse/thingweb.node-wot/issues/728 where the types for the `bl` and `lru-cache` modules could not be found during transpilation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "coap",
-  "version": "1.0.4",
+  "version": "1.0.5",
   "description": "A CoAP library for node modelled after 'http'",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "@types/chai": "^4.3.0",
     "@types/debug": "^4.1.7",
     "@types/mocha": "^9.1.0",
+    "@types/node": "^17.0.10",
     "@types/sinon": "^10.0.8",
     "@typescript-eslint/eslint-plugin": "^4.33.0",
     "chai": "^4.3.4",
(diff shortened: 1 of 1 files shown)


relu91:
I think something is off, `@type/x` is meant to be used as `devDependecies` and everything should work correctly. Because in practice you distribute only the js file and the d.ts types.

JKRhb:
Hmm, I think at least `bl` is an exception to the rule as `OutgoingMessage` extends/implements `BufferList` from `bl`. Therefore, the ts transpiler needs the type information to work. As the types are included as a devDependency here, node-wot can't access the types, though. That is the reason the ts transpiler is complaining, I think.

relu91 (approved):
Ok, I've looked around and I find out that **only in this particular case** is ok to have @types in dependencies. Sadly this will increase the bundle size for your JS-only users (without any major benefit for them). See https://stackoverflow.com/a/46011417/925955

On the other hand, I think you can get rid of  `@types/node`.

Apollon77:
so we sta with types for bl AND lru-cache in deps?

JKRhb:
Hmm, yeah, there was also an error thrown for lru-cache. I just noticed that here we have a similar case, as the `CoapLRUCache` extends `LRUCache`. So I think both have to go into the regular dependencies.

We could try removing `@types/node` by the way, not sure if anything directly depends on a definition from the module, though.

JKRhb:
I now removed `@types/node` from the regular dependencies. Transpiling in dependent projects should still work (tested out using `npm install --production`).

JKRhb:
@Apollon77 Could you release a new version after merging this PR? :)

Apollon77:
DO you want to add the version increases to the PR already? ;-)

JKRhb:
> DO you want to add the version increases to the PR already? ;-)

Sure, done :)

danielpeintner:
Is there any blocking issue left?

Apollon77:
@danielpeintner just me having time to merge and release it together with an other fix ...

PR #337, opened 2022-04-19
fix(agent): fix bug in observe option check

JKRhb:
While #330 mostly fixed the problems with the observe request parameter, it also introduced a new bug that caused the Agent to always set the observe option to 0 in requests. This PR fixes the issue, making the Agent behave normally again.


Apollon77:
Do you want to dorectly include versdion increase for release?

JKRhb:
> Do you want to dorectly include versdion increase for release?

Oh, sure, I'll update the PR :)

JKRhb:
Done :)

PR #338, opened 2022-04-23
chore: replace nyc with c8 for code coverage

JKRhb:
Having a look at [coveralls.io](https://coveralls.io/github/mcollina/node-coap), I noticed that code coverage seems to be a bit broken right now. I recently found a better tool than nyc called c8 that seems to work with V8's native code coverage capabilities and be used as a drop-in replacement, fixing typescript related issues with nyc. It also gets rid of one additional dev-dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,8 +8,8 @@
     "build": "tsc -b",
     "pretest": "npm run build",
     "prepublishOnly": "npm run build",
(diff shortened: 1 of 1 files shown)


JKRhb:
Oh, as you can see the improvement above already :) nyc wasn't capable of handling typescript, as it seems.

PR #341, opened 2022-04-23
docs: describe compatible node versions in README

JKRhb:
As we only test for node 12 and above at the moment, it probably makes sense to only test for these versions. As soon as node 12 is EOL, we should bump the version to 14.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,8 @@ If you need a command line interface for CoAP, check out
 
 **node-coap** is an **OPEN Open Source Project**, see the <a href="#contributing">Contributing</a> section to find out what this means.
 
+The library *should* run on Node versions 10 and above but is only tested with LTS versions (currently 12, 14, and 16).
+
 <a name="install"></a>
 ## Installation
 


Apollon77:
Hm ... honestly ... I personally have no issue in stating in readme for which versions we test and that older versions might work (unless we explicitely know of a breakage), and so set the minimum version to "reality" ... 

I personally hate libraries that block out users of older versions because the reality is that many users do "never change a running version". So increase version when needed - fine, but else allow others to use also older versions (at their own risk). As soon as we learn or break it with a dep then we should increase it.

But this is more a "paradigm discussion" :-)

JKRhb:
That is actually a very good point :) I guess one benefit of the typescript compilation is that we can easily target older versions. Should we add node 10 back to the CI then as well, and only add the engines entry if things should break for that version?

Apollon77:
I think nodejs 10 will break in testing because testing libs kicked it ... so as said ... I would just state in Readme which versions are tested but that also nodejs 10 should work from current knowledge

JKRhb:
I updated the PR accordingly :) Feel free to make suggestions for the wording :)

Apollon77:
Perfect

PR #343, opened 2022-04-23
fix: fix observe when receiving block2 response

### LINKED ISSUE #336: reset msg on separated response(observe) with block2 option
am facing an issue, 
1. node-coap is a client and sends a CON request with observe and block2 option.
2. serve response with ACK including block2 option with moreFlag set to false.
3. when the new data is ready at server its send a notification CON message with block2 and moreFlag=true here client sents RESET 
instead of ACK.
Am i doing something wrong here. 
export

JKRhb:
This PR incorporates the suggested change by @Thousif-khan and should fix #336.

### DIFF
--- a/lib/agent.ts
+++ b/lib/agent.ts
@@ -332,7 +332,7 @@ class Agent extends EventEmitter {
                 // TODO There is a previous response but is not an ObserveStream !
                 return
             }
-        } else if (block2 != null && packet.token != null) {
+        } else if (block2 != null && packet.token != null && !observe) {
             this._tkToReq.delete(packet.token.toString('hex'))
         } else if (!observe && !req.multicast) {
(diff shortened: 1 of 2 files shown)


JKRhb:
I think we should probably merge #344 first (which addresses some more fundamental observe issues) and see afterwards if this PR is still needed.

Apollon77:
Ok, I release a 1.0.8 with #34 in and then we test again

JKRhb:
> Ok, I release a 1.0.8 with #34 in and then we test again

Awesome, thank you :)

JKRhb:
Hmm. Not sure if the additional if check changed anything?

Apollon77:
Will merge and release tonight. Sorry for the delay

PR #347, opened 2022-05-20
chore: update tested node versions to 14, 16, and 18

JKRhb:


JKRhb:
@Apollon77 Hmm, seems like Node 18 causes some problems. I think there was a multicast-related API change that causes that?

Apollon77:
puuhh ... ist it that "network family parameter change"? Will be reverted in 18.2 as it seems

JKRhb:
> puuhh ... ist it that "network family parameter change"? Will be reverted in 18.2 as it seems

Oh, you mean this one, right? https://github.com/nodejs/node/pull/41431#discussion_r841272416

Apollon77:
yes

JKRhb:
> nope ... not fixed by 18.2

I think the number of errors reduced from 10 to 4, though, so we're at least halfway there ;)

JKRhb:
The latest version of Node 18 seems to have solved the issue :) So I would merge this PR.

PR #349, opened 2022-10-18
chore: address lint issues

Apollon77:

PR #351, opened 2022-10-26
feat(server): add Size2 logic

JKRhb:
As a follow-up to #348, this PR adds support for the Size2 option to the server, allowing clients to query the total payload size of a resource during block-wise transfer.

(CC @relu91: After this is merged and published, node-wot should fulfill this assertion by default :))


JKRhb:
I'll rebase this PR after #349 is merged :)

PR #352, opened 2022-10-27
Create FUNDING.md

Apollon77:
@mcollina Are you ok with adding also add to the sponsoring for the library?

### DIFF
--- a/.github/FUNDING.md
+++ b/.github/FUNDING.md
@@ -1,0 +1,2 @@
+github: [mcollina, Apollon77]
+custom: ['https://paypal.me/jkrhb ', 'https://paypal.me/Apollon77']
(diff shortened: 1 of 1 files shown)


JKRhb (approved):

mcollina:
Please don't include me. I don't plan to maintain this library anymore.

@Apollon77 may I move this to you?

Apollon77:
@mcollina All good, we take over :-)

Please move to https://github.com/coapjs orga ... I synced with @JKRhb and we think thats best. it is just founded orga and maybe easier for the future to manage stuff.

What do you think about the idea that we also take over coap-packet ? I think for that I would then also need npm rights ifyou think it is a good idea.

mcollina:
can you invite me to the org? I'll move both packages.

mcollina (approved):
lgtm

Apollon77:
@mcollina Thank you for your trust the last year! I'm sure we see us again in several projects ;-)

PR #366, opened 2023-05-19
fix: keep user-defined options during block2 transfers

### LINKED ISSUE #365: Missing options in COAP response
I am trying to add a `Content-Format` option to my COAP response, this seems to work when the data size is below `1280`. But if it exceeds `1280` the data is received, but removes the added option. I have checked the response in wireshark and the option is there, so it must be where the packet is reconstructed on the client.

Working example with a small packet
```
CoapServer.on('request', (req: IncomingMessage, res: OutgoingMessage) => {
    const data = "a".repeat(10);
    res.code = COAP.HttpToCoapStatus(resp.status);
    res.setOption("Content-Format", 'application/cbor');
    res.end(CBOR.encode(data));
});
```
Capture

Here I loose the `Content-Format` option on the client, see screenshot:
```
CoapServer.on('request', (req: IncomingMessage, res: OutgoingMessage) => {
    const data = "a".repeat(1300);
    res.code = COAP.HttpToCoapStatus(resp.status);
    res.setOption("Content-Format", 'application/cbor');
    res.end(CBOR.encode(data));
});
```

Capture2

I assume this is to do with when the client receiving a blockwise transfer?

JKRhb:
Fixes #365.

### DIFF
--- a/lib/server.ts
+++ b/lib/server.ts
@@ -81,6 +81,11 @@ class CoapLRUCache<K extends {}, V extends {}> extends LRUCache<K, V> {
     pruneTimer: NodeJS.Timeout
 }
 
+interface Block2CacheEntry {
+    buffer: Buffer
+    options: Option[]
+}
+
 class CoAPServer extends EventEmitter {
     _options: CoapServerOptions = {}
     _proxiedRequests: Map<string, MiddlewareParameters> = new Map()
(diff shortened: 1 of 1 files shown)


JKRhb:
> I approved, but maybe lets wait if the user can check it

Thank you :) Yeah, let's wait for them. In the meantime, I rebased the PR and squashed the commits into one.

Apollon77:
feel free to merge :-)

JKRhb:
For some reason, the CI fails on Master now for Node.js 20 :/ Will try to investigate that, maybe we need to temporarily revert the merge, though.

PR #370, opened 2023-06-04
Do not include network aliases (identified by same MAC)

htool:
Starting a listener on a network alias fails, so do not include them in allAddresses function.

### DIFF
--- a/lib/server.ts
+++ b/lib/server.ts
@@ -69,6 +69,7 @@ function allAddresses (type): string[] {
     for (const ifname in interfaces) {
         if (ifname in interfaces) {
             interfaces[ifname]?.forEach((a) => {
+                // Checking for repeating MAC address to avoid trying to listen on same interface twice
                 if (a.family === family && !macs.includes(a.mac)) {
                     addresses.push(a.address)
                     macs.push(a.mac)
--- a/lib/server.ts
+++ b/lib/server.ts
@@ -64,12 +64,14 @@ function allAddresses (type): string[] {
         family = 'IPv6'
     }
     const addresses: string[] = []
+    const macs: string[] = []
     const interfaces = os.networkInterfaces()
     for (const ifname in interfaces) {
         if (ifname in interfaces) {
             interfaces[ifname]?.forEach((a) => {
-                if (a.family === family) {
+                if (a.family === family && macs.includes(a.mac) == false) {
                     addresses.push(a.address)
+                    macs.push(a.address)
                 }
             })
         }
--- a/lib/server.ts
+++ b/lib/server.ts
@@ -69,7 +69,7 @@ function allAddresses (type): string[] {
     for (const ifname in interfaces) {
         if (ifname in interfaces) {
             interfaces[ifname]?.forEach((a) => {
-                if (a.family === family && macs.includes(a.mac) == false) {
+                if (a.family === family && !macs.includes(a.mac)) {
                     addresses.push(a.address)
                     macs.push(a.address)
                 }
--- a/lib/server.ts
+++ b/lib/server.ts
@@ -71,7 +71,7 @@ function allAddresses (type): string[] {
             interfaces[ifname]?.forEach((a) => {
                 if (a.family === family && !macs.includes(a.mac)) {
                     addresses.push(a.address)
-                    macs.push(a.address)
+                    macs.push(a.mac)
                 }
             })
         }


htool:
Solves https://github.com/coapjs/node-coap/issues/355

lib/server.ts
```diff
@@ -64,12 +64,14 @@ function allAddresses (type): string[] {
         family = 'IPv6'
     }
     const addresses: string[] = []
+    const macs: string[] = []
     const interfaces = os.networkInterfaces()
     for (const ifname in interfaces) {
         if (ifname in interfaces) {
             interfaces[ifname]?.forEach((a) => {
-                if (a.family === family) {
+                if (a.family === family && macs.includes(a.mac) == false) {
```
JKRhb, line 72: Thank you, the changes look good to me! :)

As indicated by the linter, though, this line should probably be changed as follows:

```suggestion
                if (a.family === family && !macs.includes(a.mac)) {
```

JKRhb (approved):

lib/server.ts
```diff
@@ -64,12 +64,14 @@ function allAddresses (type): string[] {
...
     const addresses: string[] = []
+    const macs: string[] = []
     const interfaces = os.networkInterfaces()
     for (const ifname in interfaces) {
         if (ifname in interfaces) {
             interfaces[ifname]?.forEach((a) => {
-                if (a.family === family) {
+                if (a.family === family && !macs.includes(a.mac)) {
                     addresses.push(a.address)
+                    macs.push(a.address)
```
Apollon77, line 74: Should this not push a.mac??

htool:
Good catch on my copy mistake.

Scenario is when having multiple ips on the same interface.
The interface can only have 1 listener.

JKRhb:
Can you maybe post a code snippet or script for reproducing the error? I tried adding an alias to a network interface (using something along the lines of `ip link set foo alias bar`) but the original code seems to still work fine afterward. It is not unlikely that I made an obvious mistake here, though.

htool:
Ip addr add to add a second ip to an interface. It's not alias but additional ip on same interface.

Apollon77:
Hm .. but why this do not work? Do we not listen on that up, so multiple listeners should work. Which error exactly you get?

htool:
Adding second ip to existing wifi interface:
$ sudo ip addr add 192.168.1.125 dev wlp64s0

allAddresses will list both IPs, while being the same interface.

```
Error: addMembership EADDRINUSE
    at Socket.addMembership (node:dgram:849:11)
    at /home/hans/.signalk/node_modules/coap/dist/lib/server.js:207:34
    at Array.forEach (<anonymous>)
    at Socket.<anonymous> (/home/hans/.signalk/node_modules/coap/dist/lib/server.js:206:58)
    at Socket.onListening (node:dgram:252:7)
    at Socket.emit (node:events:513:28)
    at startListening (node:dgram:181:10)
    at node:dgram:366:7
    at processTicksAndRejections (node:internal/process/task_queues:84:21)
    at runNextTicks (node:internal/process/task_queues:65:3)
    at processTimers (node:internal/timers:499:9) {
  errno: -98,
  code: 'EADDRINUSE',
  syscall: 'addMembership'
}
```

JKRhb:
Thank you for the stack trace! I think the problem here is actually that the server tries to join the same multicast group twice. For some reason, it iterates over the list of IP addresses instead of a list of network interfaces. Therefore, if you add another IP address (which appears to be used as an identifier here), calling `Socket.addMembership()` will fail.

Maybe we could try replacing the addresses as identifiers with `os.getNetworkInterfaces()` to get the proper interface names..? I need to have a closer look into this later.

JKRhb:
Since `allAddresses` is apparently only used for getting the addresses as interface identifiers for the multicast configuration, I think the approach proposed by @htool is actually an appropriate fix. Maybe we could rename the `allAddresses` method or add some documentation to make its (de-facto) purpose a bit clearer.

htool:
@Apollon77 What still needs to be done to move this forward? I don't see how I can further influence it.

Apollon77:
@htool Maybe see the last comment from @JKRhb as reveiw feedback:

> Maybe we could rename the allAddresses method or add some documentation to make its (de-facto) purpose a bit clearer.

And rename method or add acomment describing the behavior and why? Then I think we can merge here.

@JKRhb any other opinion?

htool:
I've added a comment to explain why the repeating MAC check is there.

JKRhb (approved):
Thank you, @htool, looks good to me :)

Unfortunately, the tests under Node 20 are still failing, but since that is a general problem that is also present on the main branch, I guess we can go forward with merging this PR. Hopefully, we'll be able to find a fix for that soon.

PR #373, opened 2023-11-19
chore: upgrade dependencies to their latest versions

JKRhb:
This PR updates all dependencies except `readable-stream` and `sinon`. Especially the latter one needs some more special treatment to keep working.


index.ts
```diff
@@ -15,7 +15,7 @@ import ObserveWriteStream from './lib/observe_write_stream'
 import { parameters, refreshTiming, defaultTiming } from './lib/parameters'
 import { isIPv6 } from 'net'
 import { registerOption, registerFormat, ignoreOption } from './lib/option_converter'
-import { CoapServerOptions, requestListener, CoapRequestParams, ParametersUpdate, AgentOptions, CoapPacket, Option, OptionValue } from './models/models'
+import { type CoapServerOptions, type requestListener, type CoapRequestParams, type ParametersUpdate, type AgentOptions, type CoapPacket, type Option, type OptionValue } from './models/models'
```
Apollon77, line 18: Would not "import type {...}" also work and be more clear?

test/server.ts
```diff
@@ -654,7 +654,7 @@ describe('server', function () {
             fastForward(10, 1000)
         })
 
-        it('should reply with a confirmable after an ack', function (done) {
+        it.skip('should reply with a confirmable after an ack', function (done) {
```
Apollon77, line 657: Why do we skip this test?

index.ts
JKRhb: Thank you :) I wasn't aware of this possibility, should now be fixed.

test/server.ts
JKRhb: That was an unintended leftover, should now be removed :)

Apollon77 (approved):

Apollon77:
Approved, merge wehen you are ready

JKRhb:
> Approved, merge wehen you are ready

Awesome, thank you :)

PR #374, opened 2023-11-19
chore: update lru-cache dependency, update Node versions in CI

JKRhb:
Spin-off from #373 that upgrades the `lru-cache`, because it is a bit tedious to debug otherwise.

Also updates the Node versions used by the CI, since the most recent `lru-cache` version seems to be incompatible with Node 14. Since Node 14 is EOL by way, applying this update might be overdue anyway.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,7 +13,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x]
+        node-version: [18.x]
         os: [ubuntu-latest]
 
     steps:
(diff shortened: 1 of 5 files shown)


Apollon77 (approved):
Al right ... makes sense, BUT you also need to change https://github.com/coapjs/node-coap/blob/master/package.json#L68 !

JKRhb:
> Al right ... makes sense, BUT you also need to change https://github.com/coapjs/node-coap/blob/master/package.json#L68 !

Thank you, good catch! :) I now set the minimum version to 18, by the way, as I remembered that 16 was phased out early due to the EOL of OpenSSL 1.1.1

JKRhb:
If the current changes still work for you, then you could just hit the merge button :)

Issue #375, opened 2023-12-14
Max packet size error when payload length close to `maxPacketSize`


edrose-tewke:
### Description

When node-coap is used as a server it is normally not required to explicitly set the `Block2` option when sending a response, even if the payload length is greater than the `maxPacketSize` setting. This is because of a check performed at [server.ts:678](https://github.com/coapjs/node-coap/blob/master/lib/server.ts#L678) which should detect that the payload requires a blockwise transfer, and then automatically apply the blockwise transfer logic to ensure that the transfer goes ahead.

Unfortunately this check is performed on just the payload, without regard for the size of the header. This means that the logic does not work for payload sizes that are close to, but not over, the `maxPacketSize` parameter. Depending on the header size, this holds for packets that are roughly between 1268 - 1280 bytes long.

### Code to Recreate

The problem can easily be recreated by modifying `examples/blockwise.js`. If the client does not add the `Block2` header (since it does not know what size the response will be), and the server generates a payload of 1268 bytes, the server will crash with the message "Max packet size is 1280: current is 1281".

If the requested payload size is increased on line 16 to a value >= 1280 then the server automatically applies blockwise transfers and no error is thrown. If it is decreased to <= 1267 then the payload doesn't need to be split and gets sent in a single response. The error only appears when (in pseudocode) `payload.length < maxPacketSize && payload.length + header.length > maxPacketSize`.

```javascript
const coap = require('../') // or coap

coap.createServer((req, res) => {
    // FIXME: This has became a bit ugly due to the
    //        replacement of the depracated url.parse
    //        with the URL constructor.
    //        This part of the exmample should be replaced
    //        once a nicer solution has been found.

    const splitURL = req.url.split('?')
    const time = parseInt(splitURL[1].split('=')[1])
    const pathname = splitURL[0].split('/')[1]

    res.end(new Array(time + 1).join(pathname)) // <-- Make it easier to specify exact size
}).listen(() => {
    const req = coap.request('coap://localhost/1?t=1268') // <-- Request a payload of 1268 bytes

    // edit this to adjust max packet
    //req.setOption('Block2', Buffer.of(0x2)) // <-- Remove client Block2 option

    req.on('response', (res) => {
        res.pipe(process.stdout)
        res.on('end', () => {
            process.exit(0)
        })
    })

    req.end()
})

```

Here is a patch that can be applied to the master branch to change the example if needed
```patch
iff --git a/examples/blockwise.js b/examples/blockwise.js
[...]

PR #376, opened 2023-12-17
Fix blockwise response logic

edrose:
## Overview

This is a fix for #375 where the server will attempt to generate a packet greater than the maximum allowable, which was set to 1280 bytes. This would occur if the size of the payload was less than 1280 bytes, but the size of the resulting CoAP message was greater than 1280 bytes after the headers were added.

The bug was introduced 9 years ago in d727d430d7c744a516d4c68d156ed6bf8262f7b7 and has some incorrect logic for dealing with message sizes. It would be very easy to simply revert those changes and call it a fix, but I wanted to try and understand what the author was trying to implement and to fix the implementation, rather than simply removing functionality.

In addition I've updated the tests to be more robust and to catch errors like this in the future.

## Motivation and Context

The intent of the above commit, I believe, was to allow the maximum packet size to be adjusted for cases where the server is running over a network stack that has an MTU different to that of ethernet. A 6LoWPAN/Thread network stack, for example, will have an MTU significantly less than that of ethernet so the server can be configured to perform blockwise transfers earlier to avoid message fragmentation at the IP layer. This is done with an `IP_MTU` constant and a `maxPacketSize` parameter that control when blockwise transfers are applied.

Unfortunately the phrase 'packet' and 'MTU' are used all over the place and don't really refer to anything meaningful in the context of the application layer. The IP MTU is the size of the whole IP packet, including all headers and payload. The `IP_MTU` constant defaults to 1280 and is used only to check the CoAP payload, missing the CoAP headers, UDP header, and IP header when performing the check.

The CoAP specification does not ever refer to anything as a 'packet' - it refers to 'messages'. A 'packet' is the IP layer 3 data unit, and I think this ambiguity of naming in the library has caused confusion in the past when features are implemented.

## Changes Introduced

Even though the maximum CoAP message size is _based_ on the IP MTU, it's not practical to try and determine the MTU for every response and apply blockwise logic accordingly. Furthermore the header size of the IP layer is not fixed, and when using 6LoWPAN neither is the UDP header size. So instead the changes here follow the recommendations set out in [RFC 7252 Section 4.6](https://datatracker.ietf.org/doc/html/rfc7252#section-4.6) which state:

> If the Path MTU is not known for a destination, an IP MTU of 1280
> bytes SHOULD be assumed; if nothing is known about the size of the
> headers, good upper bounds are 1152 bytes for the message size and
> 1024 bytes for the payload size.

Those recommendations are reflected in the defaults that are implemented, and can be adjusted from the server parameters.

### Tests

Update tests in `test/blockwise.ts` to check the response code of the server when checking responses and to move the test payload size to within the bounds of the edge case that this is fixing.

### Server Code

Change the parameters file to remove the `IP_MTU` constant. This was replaced with two constants: `MAX_PAYLOAD_SIZE` and `MAX_MESSAGE_SIZE`. Those changes are reflected in the parameters by removing the `maxPacketSize`, replacing it with `maxPayloadSize` and `maxMessageSize` respectively.

The blockwise logic has been updated to be more efficient and understandable, using the provided `maxPayloadSize` parameter. The check for whether a response needs to be split into blocks is now correct, fixing the initial bug.

The `maxMessageSize` parameter is used when calling [`coap-packet:generate()`](https://github.com/coapjs/coap-packet/blob/61ffea1059c0ee005c60bcb9e904455dc2065bdd/index.js#L31) and is passed as the `maxLength` argument. This argument, when not explicitly passed, defaults to 1280 which (as discussed) does not represent the maximum allowable message size. Now an error will be returned from that function when the total CoAP message size is greater than the max allowable set in the server. This error should now only occur if the server is misconfigured.

## Potentially Breaking Changes

The changes to the parameters might be breaking changes, depending on how they are used in other projects. I'm happy to rename them if required in order to maintain some form of backwards compatibility, at the expense of the names being less clear.

### DIFF
--- a/lib/parameters.ts
+++ b/lib/parameters.ts
@@ -131,6 +131,12 @@ const NON_LIFETIME = 145
 const COAP_PORT = 5683
 
 /**
+ * Maximum total size of the CoAP application layer message, as
+ * recommended by the CoAP specification
+ */
+const MAX_MESSAGE_SIZE = 1152
+
+/**
  * Default max payload size recommended in the CoAP specification
  * For more info see RFC 7252 Section 4.6
  */
@@ -178,6 +184,7 @@ const p: Parameters = {
     nonLifetime: NON_LIFETIME,
     coapPort: COAP_PORT,
     maxPayloadSize: MAX_PAYLOAD_SIZE,
+    maxMessageSize: MAX_MESSAGE_SIZE,
     sendAcksForNonConfirmablePackets,
     piggybackReplyMs,
     pruneTimerPeriod
--- a/lib/parameters.ts
+++ b/lib/parameters.ts
@@ -131,9 +131,10 @@ const NON_LIFETIME = 145
 const COAP_PORT = 5683
 
 /**
- * Default max packet size based on IP MTU.
+ * Default max payload size recommended in the CoAP specification
+ * For more info see RFC 7252 Section 4.6
  */
-const IP_MTU = 1280
+const MAX_PAYLOAD_SIZE = 1024
 
 /* Custom default parameters */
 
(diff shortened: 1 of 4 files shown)


JKRhb (approved):
Hi @edrose, thank you for your (very well-documented) PR, and sorry for the long response time! The changes look good to me, if you are also fine with the PR, @Apollon77, then we could merge it :)

Apollon77 (approved):

PR #397, opened 2026-02-15
feat: Native OSCORE (RFC 8613) Integration

stoprocent:
# feat: Native OSCORE (RFC 8613) Integration

## Summary

Add built-in OSCORE (Object Security for Constrained RESTful Environments) support to node-coap using the [`coap-oscore`](https://github.com/stoprocent/node-coap-oscore) package. OSCORE provides end-to-end encryption at the CoAP message level, operating on raw binary buffers so that all existing CoAP features — block transfers, observe, caching, retransmissions — work transparently over encrypted channels.

### New Files

| File | Description |
|------|-------------|
| `lib/oscore.ts` | `SecurityContextManager` — server-side registry for OSCORE contexts, keyed by recipientId + optional idContext. Manages token-to-context bindings for response encoding and aggregates SSN change events for persistence. |
| `lib/oscore_helpers.ts` | OSCORE option parsing utilities — extracts and parses the OSCORE option (number 9) from CoAP packets per RFC 8613 Section 6.1 (KID, KID Context, PIV fields). |
| `test/oscore.ts` | Comprehensive test suite with 12 tests covering all major scenarios. |

### Modified Files

| File | Changes |
|------|---------|
| `package.json` | Added `coap-oscore` dependency |
| `models/models.ts` | Extended `AgentOptions` (`oscoreOnly`), `CoapServerOptions` (`oscoreContexts`, `oscoreOnly`), and `MiddlewareParameters` (OSCORE context/metadata fields) — all optional, zero breaking changes |
| `lib/incoming_message.ts` | Added `oscoreContext` property and `isOscore` getter for server-side request introspection |
| `lib/agent.ts` | OSCORE context map with `addOscoreContext()`/`removeOscoreContext()`, async decode on incoming messages via `_handlePlainMessage()` extraction, `sender.send()` wrapping for transparent outgoing encryption, observe `_disableFiltering` for OSCORE streams, Echo auto-retry for 4.01+Echo responses |
| `lib/middlewares.ts` | New `oscoreDecryptRequest` middleware — detects OSCORE option, looks up context by KID, decodes raw buffer, re-parses inner packet, binds token for response encoding, handles Echo challenge (status 201) for first-request-after-reboot. Updated `handleServerRequest` to forward OSCORE metadata. |
| `lib/server.ts` | `_oscoreContextManager` and `_oscoreOnly` fields, middleware pipeline insertion, `_finishSendResponse()` helper extracted for async OSCORE response encoding, token lifecycle management (unbind on response/observe finish), updated `_handle()` signature to accept OSCORE metadata, dynamic `addOscoreContext()`/`removeOscoreContext()` with lazy middleware initialization |
| `index.ts` | Export `SecurityContextManager`, re-export `OSCORE`, `OscoreContext`, `OscoreContextStatus` types from `coap-oscore` |
| `README.md` | Added RFC 8613 to introduction, new OSCORE section with client/server/observe/dynamic examples, documented all new API surface (server options, agent methods, IncomingMessage properties, SecurityContextManager) |

## Design Decisions

1. **OSCORE at buffer boundaries** — Encrypt after `generate()`, decrypt before `parse()`. This keeps all existing CoAP logic working on decrypted packets with zero modifications to block transfer, observe, or caching code.

2. **Agent wraps `sender.send()`** — Instead of modifying every `generate()` call site, the agent monkey-patches `sender.send()` to encrypt all outgoing buffers. This automatically handles initial requests, block1 segments, block2 continuations, and retransmissions (RetrySend stores the already-encoded buffer).

3. **Server uses middleware pipeline** — `oscoreDecryptRequest` slots into the existing fastseries middleware chain between `parseRequest` and `handleServerRequest`. Uses `.then()/.catch()` (not async/await) for fastseries compatibility.

4. **Per-block OSCORE** — Per RFC 8613 §4.1.3.4, each block-wise message is independently OSCORE-protected. The sender.send() wrapping handles this automatically.

5. **All additions are optional** — Every new field is optional. A server/agent without OSCORE configuration behaves identically to before.

## Test Coverage

| Test | Description |
|------|-------------|
| GET round-trip | Basic GET with OSCORE encryption/decryption, verifies `isOscore` and `oscoreContext.senderId` |
| POST round-trip | POST with payload, OSCORE-protected in both directions |
| Mixed secure/insecure | Server handles both OSCORE and plaintext requests correctly |
| oscoreOnly server | Server rejects plaintext with 4.01 |
| oscoreOnly agent | Agent throws on requests to peers without context |
| Multiple contexts on agent | Separate OSCORE contexts route correctly to different peers |
| Multiple contexts on server | Server handles requests from different OSCORE clients |
| Unknown client | Server returns 4.01 for OSCORE request with unregistered KID |
| SSN persistence | Verifies `ssn` events fire via SecurityContextManager |
| Dynamic add/remove (agent) | Add and remove contexts at runtime |
| Dynamic add/remove (server) | Add context at runtime on server created without OSCORE |
| Observe with OSCORE | Observe notifications encrypted transparently, `_disableFiltering` active |

**Results: 486 passing (474 existing + 12 new), 6 pending (unchanged), 0 failing**

### DIFF
--- a/lib/agent.ts
+++ b/lib/agent.ts
@@ -110,7 +110,8 @@ class Agent extends EventEmitter {
                         }
                     })
                     .catch(() => {
-                        this._handlePlainMessage(msg, rsinfo)
+                        // OSCORE decode failed — drop the message.
+                        // Do NOT fall back to plaintext when a security context exists.
                     })
                 return
             }
@@ -376,13 +377,41 @@ class Agent extends EventEmitter {
         if (packet.code === '4.01' && req != null && this._getOscoreContext(rsinfo.address, rsinfo.port) != null) {
             const echoOpt = getOption(packet.options, '252')
             if (echoOpt != null) {
-                const retryUrl = { ...req.url, token: req._packet.token }
-                const retryReq = this.request(retryUrl)
-                retryReq.setOption('252', echoOpt)
-                retryReq.on('response', (res) => req.emit('response', res))
-                retryReq.on('error', (err) => req.emit('error', err))
-                retryReq.end()
-                return
+                // Limit Echo retries to prevent infinite loops from malicious servers
+                const retryCount = (req as any)._echoRetries ?? 0
+                if (retryCount >= 1) {
+                    // Max retries exceeded — deliver the 4.01 to the application
+                    // (fall through to normal response handling below)
+                } else {
+                    const retryUrl = { ...req.url, token: req._packet.token }
+                    const retryReq = this.request(retryUrl)
+
+                    // Carry over options from original packet (Content-Format, Accept, custom, etc.)
+                    // Skip Uri-Path, Uri-Query, Observe — these are reconstructed from retryUrl by request()
+                    const skipOptions = new Set(['Uri-Path', 'Uri-Query', 'Observe'])
+                    if (req._packet.options != null) {
+                        for (const opt of req._packet.options) {
+                            if (!skipOptions.has(String(opt.name))) {
+                                retryReq.setOption(String(opt.name), opt.value)
+                            }
+                        }
+                    }
+                    retryReq.setOption('252', echoOpt)
+
+                    ;(retryReq as any)._echoRetries = retryCount + 1
+
+                    retryReq.on('response', (res) => req.emit('response', res))
+                    retryReq.on('error', (err) => req.emit('error', err))
+
+                    // Carry over payload from original request
+                    const payload = req.slice()
+                    if (payload.length > 0) {
+                        retryReq.end(payload)
+                    } else {
+                        retryReq.end()
+                    }
+                    return
+                }
             }
         }
 
--- a/lib/agent.ts
+++ b/lib/agent.ts
@@ -11,6 +11,7 @@ import { Socket, createSocket } from 'dgram'
 import { AgentOptions, CoapRequestParams, Block } from '../models/models'
 import { EventEmitter } from 'events'
 import { parse, generate, ParsedPacket } from 'coap-packet'
+import type { OSCORE } from 'coap-oscore'
 import IncomingMessage from './incoming_message'
 import OutgoingMessage from './outgoing_message'
 import ObserveStream from './observe_read_stream'
@@ -35,6 +36,8 @@ class Agent extends EventEmitter {
     _lastMessageId: number
     private _msgInFlight: number
     _requests: number
+    private _oscoreContexts: Map<string, OSCORE>
+    private _oscoreOnly: boolean
     constructor (opts?: AgentOptions) {
         super()
 
@@ -53,8 +56,23 @@ class Agent extends EventEmitter {
         }
 
         this._opts = opts
+        this._oscoreContexts = new Map()
+        this._oscoreOnly = opts.oscoreOnly ?? false
 
         this._init(opts.socket)
+    }
+
+    addOscoreContext (host: string, port: number, instance: OSCORE): void {
+        const key = `${host}:${port}`
+        this._oscoreContexts.set(key, instance)
+    }
+
+    removeOscoreContext (host: string, port: number): void {
+        this._oscoreContexts.delete(`${host}:${port}`)
+    }
+
+    private _getOscoreContext (host: string, port: number): OSCORE | undefined {
+        return this._oscoreContexts.get(`${host}:${port}`)
     }
 
     _init (socket?: Socket): void {
@@ -66,22 +84,29 @@ class Agent extends EventEmitter {
 
         this._sock = socket ?? createSocket({ type: this._opts.type ?? 'udp4' })
         this._sock.on('message', (msg, rsinfo) => {
-            let packet: ParsedPacket
-            try {
-                packet = parse(msg)
-            } catch (err) {
+            const oscoreCtx = this._getOscoreContext(rsinfo.address, rsinfo.port)
+            if (oscoreCtx != null) {
+                oscoreCtx.decode(msg)
+                    .then((decoded) => {
+                        let packet: ParsedPacket
+                        try {
+                            packet = parse(decoded)
+                        } catch {
+                            return
+                        }
+                        if (packet.code[0] === '0' && packet.code !== '0.00') {
+                            return
+                        }
+                        if (this._sock != null) {
+                            this._handle(packet, rsinfo, this._sock.address())
+                        }
+                    })
+                    .catch(() => {
+                        this._handlePlainMessage(msg, rsinfo)
+                    })
                 return
             }
-
-            if (packet.code[0] === '0' && packet.code !== '0.00') {
-                // ignore this packet since it's not a response.
-                return
-            }
-
-            if (this._sock != null) {
-                const outSocket = this._sock.address()
-                this._handle(packet, rsinfo, outSocket)
-            }
+            this._handlePlainMessage(msg, rsinfo)
         })
 
         if (this._opts.port != null) {
@@ -101,6 +126,24 @@ class Agent extends EventEmitter {
 
         this._msgInFlight = 0
         this._requests = 0
+    }
+
+    private _handlePlainMessage (msg: Buffer, rsinfo: AddressInfo): void {
+        let packet: ParsedPacket
+        try {
+            packet = parse(msg)
+        } catch (err) {
+            return
+        }
+
+        if (packet.code[0] === '0' && packet.code !== '0.00') {
+            return
+        }
+
+        if (this._sock != null) {
+            const outSocket = this._sock.address()
+            this._handle(packet, rsinfo, outSocket)
+        }
     }
 
     close (done?: (err?: Error) => void): this {
@@ -321,6 +364,20 @@ class Agent extends EventEmitter {
             }
         }
 
+        // Echo auto-retry for OSCORE peers
+        if (packet.code === '4.01' && req != null && this._getOscoreContext(rsinfo.address, rsinfo.port) != null) {
+            const echoOpt = getOption(packet.options, '252')
+            if (echoOpt != null) {
+                const retryUrl = { ...req.url, token: req._packet.token }
+                const retryReq = this.request(retryUrl)
+                retryReq.setOption('252', echoOpt)
+                retryReq.on('response', (res) => req.emit('response', res))
+                retryReq.on('error', (err) => req.emit('error', err))
+                retryReq.end()
+                return
+            }
+        }
+
         const observe = req.url.observe != null && [true, 0, '0'].includes(req.url.observe)
 
         if (req.response != null) {
@@ -342,6 +399,9 @@ class Agent extends EventEmitter {
 
         if (observe && packet.code !== '4.04') {
             response = new ObserveStream(packet, rsinfo, outSocket)
+            if (this._getOscoreContext(rsinfo.address, rsinfo.port) != null) {
+                (response as ObserveStream)._disableFiltering = true
+            }
             response.on('close', () => {
                 this._tkToReq.delete(packet.token.toString('hex'))
                 this._cleanUp()
@@ -452,6 +512,21 @@ class Agent extends EventEmitter {
         })
 
         req.sender = new RetrySend(this._sock, port, host, url.retrySend)
+
+        const oscoreCtx = this._getOscoreContext(host ?? '', port)
+
+        if (oscoreCtx == null && this._oscoreOnly) {
+            throw new Error('No OSCORE context for ' + (host ?? '') + ':' + port)
+        }
+
+        if (oscoreCtx != null) {
+            const originalSend = req.sender.send.bind(req.sender)
+            req.sender.send = (message: Buffer, avoidBackoff?: boolean) => {
+                oscoreCtx.encode(message)
+                    .then((encoded) => originalSend(encoded, avoidBackoff))
+                    .catch((err) => req.emit('error', err))
+            }
+        }
 
         req.url = url
 
--- a/lib/agent.ts
+++ b/lib/agent.ts
@@ -8,7 +8,7 @@
 
 import crypto = require('crypto')
 import { Socket, createSocket } from 'dgram'
-import { AgentOptions, CoapRequestParams, Block } from '../models/models'
+import { AgentOptions, CoapRequestParams, Block, Parameters } from '../models/models'
 import { EventEmitter } from 'events'
 import { parse, generate, ParsedPacket } from 'coap-packet'
 import type { OSCORE } from 'coap-oscore'
@@ -20,7 +20,7 @@ import { parseBlock2, createBlock2, getOption, removeOption } from './helpers'
 import { SegmentedTransmission } from './segmentation'
 import { parseBlockOption } from './block'
 import { AddressInfo } from 'net'
-import { parameters } from './parameters'
+import { parameters, createParameters } from './parameters'
 
 interface OscoreRsinfo extends AddressInfo {
     oscore?: boolean
@@ -42,12 +42,15 @@ class Agent extends EventEmitter {
     _requests: number
     private _oscoreContexts: Map<string, OSCORE>
     private _oscoreOnly: boolean
+    _parameters: Parameters
     constructor (opts?: AgentOptions) {
         super()
 
         if (opts == null) {
             opts = {}
         }
+
+        this._parameters = createParameters(opts.parameters)
 
         if (opts.type == null) {
             opts.type = 'udp4'
@@ -501,7 +504,7 @@ class Agent extends EventEmitter {
         const options = url.options ?? url.headers
         const multicastTimeout = url.multicastTimeout != null ? url.multicastTimeout : 20000
         const host = url.hostname ?? url.host
-        const port = url.port ?? parameters.coapPort
+        const port = url.port ?? this._parameters.coapPort
 
         const req = new OutgoingMessage({}, (req, packet) => {
             if (url.confirmable !== false) {
@@ -554,7 +557,7 @@ class Agent extends EventEmitter {
             }
         })
 
-        req.sender = new RetrySend(this._sock, port, host, url.retrySend)
+        req.sender = new RetrySend(this._sock, port, host, url.retrySend, this._parameters)
 
         const oscoreCtx = this._getOscoreContext(host ?? '', port)
 
--- a/lib/agent.ts
+++ b/lib/agent.ts
@@ -21,6 +21,7 @@ import { SegmentedTransmission } from './segmentation'
 import { parseBlockOption } from './block'
 import { AddressInfo } from 'net'
 import { parameters, createParameters } from './parameters'
+import { hasOscoreOption } from './oscore_helpers'
 
 interface OscoreRsinfo extends AddressInfo {
     oscore?: boolean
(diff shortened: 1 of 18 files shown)


stoprocent:
Hey, is there any plan to review and merge? If not I will do a `@stoprocent/node-coap` but i rather contribute here :)

stoprocent:
Hey guys, if you cannot review I will move to new namespace on npm. Just please let me know I understand everyones busy. @JKRhb @Apollon77

JKRhb:
Hi @stoprocent, sorry for getting back to you earlier! I will probably have the time to do a review by tomorrow, I hope that works for you :) In any case thank you for working on this feature, having an OSCORE implementation in node-coap is pretty exciting! :)

stoprocent:
No worries! I don’t want to rush the review. That’s the last thing I’d want to do. I just wanted to at least get it started.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "coap",
-  "version": "1.4.2",
+  "version": "1.6.0",
```
JKRhb, line 3: I think the version should only be bumped after merging this PR.

lib/agent.ts
```diff
@@ -35,13 +40,18 @@ class Agent extends EventEmitter {
     _lastMessageId: number
     private _msgInFlight: number
     _requests: number
+    private _oscoreContexts: Map<string, OSCORE>
+    private _oscoreOnly: boolean
+    _parameters: Parameters
```
JKRhb: Could this field also be `private`?

lib/agent.ts
```diff
@@ -321,13 +376,57 @@ class Agent extends EventEmitter {
...
+                    if (req._packet.options != null) {
+                        for (const opt of req._packet.options) {
+                            if (!skipOptions.has(String(opt.name))) {
+                                retryReq.setOption(String(opt.name), opt.value)
+                            }
+                        }
+                    }
+                    retryReq.setOption('252', echoOpt)
+
+                    ;(retryReq as any)._echoRetries = retryCount + 1
```
JKRhb: Do we need the `;` here?

```suggestion
                    (retryReq as any)._echoRetries = retryCount + 1
```

lib/agent.ts
```diff
@@ -342,6 +441,10 @@ class Agent extends EventEmitter {
 
         if (observe && packet.code !== '4.04') {
             response = new ObserveStream(packet, rsinfo, outSocket)
+            if (rsinfo.oscore === true) {
+                (response as ObserveStream)._disableFiltering = true
+                ;(response as ObserveStream).oscoreProtected = true
```
JKRhb: Same here?

```suggestion
                (response as ObserveStream).oscoreProtected = true
```

lib/agent.ts
```diff
@@ -451,7 +564,22 @@ class Agent extends EventEmitter {
             }
         })
 
-        req.sender = new RetrySend(this._sock, port, host, url.retrySend)
+        req.sender = new RetrySend(this._sock, port, host, url.retrySend, this._parameters)
+
+        const oscoreCtx = this._getOscoreContext(host ?? '', port)
+
+        if (oscoreCtx == null && this._oscoreOnly) {
+            throw new Error('No OSCORE context for ' + (host ?? '') + ':' + port)
```
JKRhb: This is very nitpicky, but if `host` is `undefined`, I guess the colon should not be included with the port?

lib/retry_send.ts
```diff
@@ -31,19 +32,22 @@ export default class RetrySend extends EventEmitter {
     _message: Buffer
     _timer: NodeJS.Timeout
     _bOffTimer: NodeJS.Timeout
-    constructor (sock: any, port: number, host?: string, maxRetransmit?: number) {
+    _parameters: Parameters
```
JKRhb: Could this be `private`?

JKRhb (commented):
Sorry for the delay! See a couple of initial comments below, I will continue the review in the upcoming days.

JKRhb:
Seems like the CI is failing now since the semicolons were necessary after all :/ Sorry for the mistake, could you revert these two particular changes (or find a different solution that does not involve the semicolons)? Thank you!

Apollon77:
@stoprocent Ok I also had a look and basically here is my summary:

## Overview

A 1572-line addition that wires OSCORE end-to-end encryption into node-coap by using the `coap-oscore` package as the crypto engine. The design encrypts after `generate()` and decrypts before `parse()`, so existing block / observe / cache code stays untouched. New `SecurityContextManager` (server) and `addOscoreContext()`/`removeOscoreContext()` methods (server + agent) provide the public API. All new options are optional → no breaking change. CI is green; the only previously raised review comment still open is the `package.json` version bump.

## Things that should change

### 1. `package.json` version bump (already requested)
Master is `1.4.2`; PR sets it to `1.6.0`. Per @JKRhb's existing review comment, the maintainer would do the bump on merge. Please revert `package.json:3` back to `1.4.2`.

### 2. `oscoreOnly` is silently dropped when contexts are added dynamically
In `lib/server.ts:155-159`, `_oscoreOnly` is only assigned inside the `if (this._options.oscoreContexts != null)` branch. If a user does `createServer({ oscoreOnly: true })` without `oscoreContexts` and later calls `server.addOscoreContext(...)`, the flag is ignored. Either:
- Set `this._oscoreOnly = this._options.oscoreOnly ?? false` unconditionally in the constructor, or
- Have `addOscoreContext()` also honour `this._options.oscoreOnly`.

### 3. Tight coupling to an unreleased `coap-oscore` API
`lib/middlewares.ts:201` calls `(oscore as any).clearRebootRecovery()` with the comment _"Note: clearRebootRecovery() is added in the concurrent node-oscore update"_. `package.json` pins `coap-oscore: ^2.2.1`. If a consumer resolves to a 2.x version that doesn't yet expose this method, this throws inside `.then()` and the Echo path fails silently. Either:
- Bump the dep to a `>=` that actually contains `clearRebootRecovery`, or
- Guard with `typeof oscore.clearRebootRecovery === 'function'`.

### 4. `oscoreProtected` is stamped onto streams without being declared
`lib/agent.ts:471, 490` and `lib/server.ts` set `response.oscoreProtected = true` on `ObserveStream`/`IncomingMessage` instances, but the field exists only on `IncomingMessage`. The agent code reads it back through an `as any` cast (`agent.ts:449,452`). Add `oscoreProtected: boolean` to `ObserveStream` (it already has `_disableFiltering` declared) so the typing is honest and the `as any` casts can go.

### 5. Plaintext responses still flow through to an `oscoreOnly` agent
`Agent._oscoreOnly` only blocks outgoing plaintext requests in `request()` (`agent.ts:596-599`). Incoming plaintext from a peer that has no registered context falls into `_handlePlainMessage()` and is processed normally. Either reject inbound plaintext when `_oscoreOnly === true`, or document that the flag is outbound-only.

### 6. `_pendingEchoNonces` map can grow without bound
`SecurityContextManager.storePendingEcho()` stores one nonce per `(recipientId, idContext)` and never expires it. A peer that triggers Echo and never replies leaves the entry forever. Add a TTL (analogous to `EXCHANGE_LIFETIME`) or evict on next request from that peer.

### 7. PIV length validation
`lib/oscore_helpers.ts:56` masks the low 3 bits of the flags byte and uses any value 0–7 as PIV length. Per RFC 8613 §6.1 only 0–5 are valid; 6 and 7 (and the reserved high bits 0x20/0x40/0x80) should be rejected with a malformed-option error.

### 8. Dead export
`hasOscoreOption()` in `lib/oscore_helpers.ts:90` is exported but never used. Remove or use.

## Smaller / nitpick

- **`lib/middlewares.ts:182`** — `getOption(innerPacket.options, '252' as any)` and `setOption('252', echoOpt)` (`agent.ts:427`) leak the magic option number. Define a named constant (or use the registered `'Echo'` option name) — both occurrences cast to `any` to bypass TS, which is a smell.
- **Token bookkeeping**: `bindToken()` evicts the oldest entry when at 10k cap, but a long-running observe that was bound earliest will silently lose its context binding before its `close` handler unbinds. Unlikely in practice, but worth a comment in `oscore.ts:74`.
- **`_oscoreContextManager` is `null` initially + lazily created in `addOscoreContext()`** — fine, but in `_finishSendResponse` and the encode path you use `this._oscoreContextManager?.unbindToken(...)`. The `tokenHex != null` is checked but the inner branch already constrains `tokenHex` to a non-empty string at the call site; the duplicated check is harmless but noisy.
- **Tests use bare `setTimeout(done, 500)` and `setTimeout(..., 20)` in the SSN-persistence and unknown-client tests** — these are likely to flake under CI load. Prefer event-driven completion or longer waits.
- **No test for the server-side Echo (4.01 + Echo) RFC 9175 path** — the most security-sensitive new code in `middlewares.ts:172-247` is uncovered (Coveralls confirms 42% on `middlewares.ts`).
- **README, server example**: `const { OSCORE, SecurityContextManager } = require('coap')` then `const { OscoreContextStatus } = require('coap-oscore')` — `OscoreContextStatus` is also re-exported from `coap` (per `index.ts:112`). Consistent with the import line above it would be friendlier.
- **`lib/agent.ts:9`** still uses `import crypto = require('crypto')` while other files use ES `import` syntax — minor style drift.

## Things I think are good

- Clean buffer-boundary design: encrypt in `sender.send()` wrapper, decrypt in middleware. Block / observe / RetrySend genuinely unaffected.
- Defensive crypto choices: `timingSafeEqual` for Echo comparison, namespaced `senderId:tokenHex` keying to avoid collisions across peers, refusal to fall back to plaintext when an OSCORE context exists for the peer.
- `_handlePlainMessage()` extraction and `_finishSendResponse()` extraction are good seams with no behavioural change.
- The dynamic `addOscoreContext()`/`removeOscoreContext()` API and the lazy middleware insertion are nicely done.
- Test scope (12 tests covering round-trip, mixed, oscoreOnly, multi-context, dynamic add/remove, observe, SSN) is reasonable for the surface area.

## Risk summary

| Risk | Severity |
|---|---|
| Version bump prevents merge as-is | High (blocking, easy fix) |
| `clearRebootRecovery` coupling to in-flight coap-oscore release | Medium (silent Echo failure) |
| `oscoreOnly` silently dropped without `oscoreContexts` | Medium (confusing) |
| Echo nonce map unbounded growth | Low/Medium (slow leak) |
| PIV length not validated | Low (malformed packets accepted at parse step) |
| Inbound plaintext accepted in `oscoreOnly` agent | Low (semantic) |
| Echo server-side path uncovered by tests | Medium (security-relevant code path) |

Overall this is a substantial, well-structured contribution that's close to mergeable. Addressing items 1–4 above before merge would be my recommendation; the rest can land as follow-ups.

---
_Generated by [Claude Code](https://claude.ai/code)_

Apollon77:
@stoprocent should I remove the package.json or do you want to fix that

Apollon77:
Looked at `a84ee04` (`fix(agent): only attempt OSCORE decode when outer packet carries OSCORE option`). That commit is a solid fix in its own right — RFC 8613 §4.2 forbids OSCORE-protecting empty messages, and the previous code was feeding empty ACK/RST into `oscoreCtx.decode()` and tripping the freshly-added error path on legitimate transport-layer traffic. Dropping the messageId fallback in favour of token-only matching is also the right call (prevents stray packets hijacking unrelated in-flight requests). The 8 new tests around this area (empty ACK / plain RST / NON / plaintext error / corrupted MAC / garbage / piggybacked / no-context happy path) are a nice coverage bump.

It also incidentally resolves one open item from my earlier review:

- ✅ **#8 (dead export `hasOscoreOption`)** — now used in `lib/agent.ts:96`.

### Still open from my earlier comment

1. ~~`package.json` version bump~~ — acknowledged, skipping.
2. ❌ **`oscoreOnly` silently dropped without `oscoreContexts`** — `lib/server.ts:155-159` still gates `this._oscoreOnly = …` inside the `if (this._options.oscoreContexts != null)` branch. `createServer({ oscoreOnly: true })` followed by `server.addOscoreContext(…)` still loses the flag.
3. ❌ **`clearRebootRecovery` coupling** — `lib/middlewares.ts:200-201` still has the `(oscore as any).clearRebootRecovery()` cast with the "added in the concurrent node-oscore update" comment, and `package.json` still pins `coap-oscore: ^2.2.1`. Either bump the dep floor or guard with `typeof`.
4. ❌ **`oscoreProtected` stamped without being declared on `ObserveStream`** — `lib/incoming_message.ts:32` has it, but `observe_read_stream.ts` still doesn't, so agent code still casts through `as any` to read it.
5. ⚠ **Partial — plaintext on OSCORE channel** — the new commit *does* surface plaintext responses from a registered OSCORE peer as `error` events on the matching request (great). But the general `Agent._oscoreOnly` semantics are unchanged: a peer with no registered context that sends plaintext still flows through `_handlePlainMessage()`. Either explicitly document "outbound-only" or extend the new check to apply when `_oscoreOnly === true`.
6. ❌ **`_pendingEchoNonces` unbounded growth** — `lib/oscore.ts:104` still has no TTL/eviction.
7. ❌ **PIV length validation** — `lib/oscore_helpers.ts:56` still accepts `pivLen ∈ {6,7}` (RFC 8613 §6.1 only allows 0–5), and the reserved high flag bits are not checked.

### Nitpicks still open
- Magic option number `'252'` still appears in 4 places (`agent.ts:423,444`, `middlewares.ts:182,230`) with two `as any` casts. A named constant or registering the `'Echo'` option would clean this up.
- `bindToken()` LRU eviction comment not added.
- Several `setTimeout(done, …)` patterns in `test/oscore.ts` (lines 414, 441, 457, 565, 620, 925) — likely to flake under load.
- **Server-side Echo (4.01 + Echo) RFC 9175 path is still untested** — `middlewares.ts:172-247` is the most security-sensitive new code and the new tests are all agent-side.
- README server example (`README.md:182`) still does `const { OscoreContextStatus } = require('coap-oscore')` while `OscoreContextStatus` is re-exported from `coap` itself.
- `lib/agent.ts:9` still uses `import crypto = require('crypto')` while other new files use ESM imports.

Net: one item closed, items 2–4 and 6–7 are the ones I'd most like to see addressed before merge. The fix in this commit is good and the underlying bug it closes is real.

---
_Generated by [Claude Code](https://claude.ai/code)_

Apollon77:
I will push an update soon

Apollon77:
@stoprocent I went ahead and applied the remaining open items from my reviews on `coapjs/node-coap` branch [`claude/review-pr-397-LiN3v`](https://github.com/coapjs/node-coap/tree/claude/review-pr-397-LiN3v) (commit [`21050de`](https://github.com/coapjs/node-coap/commit/21050de)). Cherry-pick or pull at your convenience.

### What's in the commit

| Item | File |
|---|---|
| Hoist `_oscoreOnly` out of the `oscoreContexts != null` gate so the flag is honoured when contexts are added at runtime | `lib/server.ts` |
| Drop unsolicited plaintext in `oscoreOnly` agent mode from peers without a registered context | `lib/agent.ts` |
| Remove `(oscore as any).clearRebootRecovery()` cast — the method is publicly typed in `coap-oscore` ≥ 2.2.1 | `lib/middlewares.ts` |
| Wrap `parseOscoreOption()` in try/catch so malformed options surface as 4.01 instead of bubbling up as 5.00 | `lib/middlewares.ts` |
| Reject reserved high flag bits (`0xe0`) and PIV lengths 6 and 7 per RFC 8613 §6.1 | `lib/oscore_helpers.ts` |
| 30 s TTL on `_pendingEchoNonces` entries so peers that never reply to a challenge can't accumulate | `lib/oscore.ts` |
| Introduce `ECHO_OPTION` constant for the magic `'252'` (RFC 9175) and use it everywhere | `lib/oscore_helpers.ts`, `lib/agent.ts`, `lib/middlewares.ts` |
| `import crypto = require('crypto')` → `import crypto from 'crypto'` (consistency) | `lib/agent.ts` |
| Collapse the README server example into one `require('coap')` line | `README.md` |
| Revert version `1.6.0 → 1.4.2`; bump `coap-oscore` floor to `^2.2.2` | `package.json` |

### New tests
- `oscoreOnly` applied via `addOscoreContext()` at runtime → 4.01 on plaintext.
- `oscoreOnly` agent silently drops unsolicited inbound plaintext.
- Malformed OSCORE option (reserved PIV length) → 4.01.
- Pending Echo nonce evicted after TTL.

24/24 OSCORE tests pass locally; 492 in the wider suite (the IPv6 hooks fail in my sandbox env but are unrelated to this change — they also fail on master).

Server-side Echo (4.01 + Echo) RFC 9175 path test that I flagged earlier is the one open coverage gap I did *not* attempt to add (it needs orchestrating an OSCORE peer that triggers `FIRST_REQUEST_AFTER_REBOOT` from `coap-oscore`, which is more involved). Probably worth a follow-up.

---
_Generated by [Claude Code](https://claude.ai/code)_

stoprocent:
Let me have a look.

package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "coap",
-  "version": "1.4.2",
+  "name": "@ts-art/node-coap",
```
Apollon77: @stoprocent i think now some unplanned changes made it in here?

Apollon77 (changes requested):

Apollon77:
@stoprocent do you plan updates or how we should best continue?

stoprocent:
@Apollon77 @JKRhb thanks again for the thorough reviews 🙏

Pushed a fix for the blocking item:

- **`package.json`** — reverted the unplanned changes. `name` is back to `coap` and `version` back to `1.4.2` (these had leaked in from my downstream release branch). Happy for you to bump the version on merge as you prefer. The only added dependency is `coap-oscore` (the OSCORE engine).

For reference, the earlier feedback was already addressed in prior commits:

- @Apollon77's items 1–8 — `oscoreOnly` set unconditionally, `clearRebootRecovery()` guarded with a `typeof` check, `oscoreProtected` declared on the stream type, inbound plaintext dropped when `oscoreOnly`, `_pendingEchoNonces` given a TTL + LRU cap, PIV length / reserved-flag validation per RFC 8613 §6.1, and `hasOscoreOption()` is now actually used. Regression tests added for the security-sensitive ones.
- @JKRhb's inline notes — the `private` fields, the `host:`-colon message, and the semicolon changes (which turned out to be necessary for ASI, so they were restored — thanks for catching the CI break).

Two honest notes:

1. A few non-OSCORE fixes rode along on this branch while it was open: a `retry_send` exponential-backoff reset bug (which also exists upstream), a `block` progress event, and per-instance CoAP parameters. If you'd rather keep this PR strictly OSCORE, I'm happy to split those into separate PRs — just say the word.
2. The server-side Echo / RFC 9175 reboot-recovery test is still `describe.skip` because it needs `coap-oscore` to expose the request-interaction upsert before encoding the 4.01+Echo challenge. I maintain `coap-oscore` too, so I can land that and flip the test on if you want it in-scope here.

Re-requesting review whenever you have a moment — no rush. Thanks!

stoprocent:
Gentle bump 🙂 @Apollon77 @JKRhb

All the requested changes have been addressed — the `package.json` name/version is reverted, and every item from the earlier reviews is in. CI is green and the branch is mergeable with no conflicts, so from my side this is ready to merge whenever you have a moment.

For what it's worth on stability: we run this build in production day to day at a fairly large scale — millions of requests daily, with close to 100 developers building against it. If there were compatibility issues I'd expect them to have surfaced by now, and so far it's been solid.

Thanks again for the reviews!

JKRhb (approved):
Hi @stoprocent, thank you for addressing the review comments and for your patience! I think we can now proceed to merging – maybe you could have a final look at the PR, @Apollon77? If the changes look good to you, feel free to hit the merge button :) Afterward we can release a new version `1.5.0`.

Apollon77:
Awesome, fine with it too ... merged

JKRhb:
A new version 1.6.0 has now been released with the changes included in this PR :) Initially, I wanted to create a release 1.5.0, but while doing so, I noticed that a release with that version number had already been published. For some reason, the commit with the updated version did not end up on the `master` branch, though.

I therefore pulled the commit back into the commit history, rearranging it in the process, before publishing the new version. That way, the code published on npm is consistent again with the corresponding tags in this repository, hoping that the changed hashes of this PR's merge commit (4ca468ce06e1b03d34636e3f937aeeb708783046) and commit for version 1.5.0 (137baaa561594b1dfb555d5e61df2999cbfbf429) will not cause any trouble. But I assume that we will be fine.

Going forward, we really need to automate the release process to ensure greater consistency between GitHub and npm. But @Apollon77 already mentioned to me that he has solution that can soon be adopted for this repository as well :)

